PR #38, opened 2015-12-26
Improve textarea helper

suzuki:
porting from '[esarea](https://github.com/fukayatsu/esarea)'.

markdown

### key bindings

- `Tab` / `Shift + Tab`
  - indent / unindent
    - multiline supported
  - move to next/prev cell in table
- `Enter` on the end of line
  - suggest list
  - suggest table
    - `| foo | bar |<hit Enter here!>`
- `Alt + Shift + Space`
  - toggle tasklist status of current line
    - tasklist: `- [x] foo`


sotarok:
great!!!!

PR #40, opened 2016-01-03
Remove bower

sotarok:
* To remove bower dependencies, use npm instead.
* If the package doesn't have package.json, add JS file to this repository (e.g. jquery.selection)
* Use browserify to require the package and use on browser.


sotarok:
To remove bower completely, we have to consider about bootstrap-sass.
We have 2 options:

* Move bootstrap 4 (alpha-2) and use npm.
    * + bootstrap 4 uses scss officially.
    * - bootstrap 4 is still alpha.
* Use bower until bootstrap 4 will be released.
    * + bootstrap migration cost is 0
    * - bower will alive.

sotarok:
oops ... there is https://github.com/twbs/bootstrap-sass and they provides their package on npm.
so use it.

PR #45, opened 2016-01-12
use MongoLab with Heroku Button

takahashim:
Compose MongoDB (mongohq) is $18/mo, so it's not good to try this application, I guess.
How about to use MongoLab instead of it?

### DIFF
--- a/app.json
+++ b/app.json
@@ -20,7 +20,7 @@
     }
   },
   "addons": [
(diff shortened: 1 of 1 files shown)


sotarok:
Thank you for the patch.
That's ... absolutely right. Thank you.

takahashim:
Thanks!

PR #75, opened 2016-04-29
Change behavior of applicationNotInstalled middleware

nukosuke:
If crowi already installed,

### In case of admin user
- redirect to `/admin`
- display error message

[https://gyazo.com/9888954b0ae608d359406badc687f27e](https://gyazo.com/9888954b0ae608d359406badc687f27e)

-----

### In case of no admin user
- redirect to `/`


sotarok:
Thank you for the proposal.
I reviewed the behaviour on review app. It's perfect!
Thank you so much!

PR #79, opened 2016-06-05
Fix error when path has only numeric chars

suzuki:
## Before

- Error at `111` line.

## After

- Fixed

### DIFF
--- a/resource/js/crowi.js
+++ b/resource/js/crowi.js
@@ -317,10 +317,10 @@ $(function() {
     var $link = $(this);
     var text = $link.text();
     var path = $link.data('path');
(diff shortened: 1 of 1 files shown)


resource/js/crowi.js
```diff
@@ -320,7 +320,7 @@ $(function() {
     var shortPath = $link.data('short-path');
```
sotarok: I think, to fix this problem, `shortPath` should be converted the return value of `data()` to string here.

suzuki: OK. Cast here.

sotarok:
Thanks!

PR #87, opened 2016-06-12
Add scroll sync between editor and preview

suzuki:
## About 

- Add scroll sync between editor and preview.
    - When the editor scrolled, the preview scrolled too.
    - When the preview scrolled, the editor did not scroll.

## Screencast

crowi4


sotarok:
Thank you for the super amazing feature!

PR #98, opened 2016-06-30
Feature search hash

sotarok:
- Change hash handling


sotarok:
Failing deploy task is maybe the problem of Heroku.

PR #100, opened 2016-07-16
Fix pagename problem

kazsw:
The page named like `/meeting`cannot be created.


sotarok:
Thank you for the feedback!
I added tests and fix some other cases, and merged this change by #105.

Thank you so much.

PR #113, opened 2016-07-31
Render and api access refactor

sotarok:
- Create `Crowi` class
- Create `CrowiRendere` class


resource/js/components/SearchPage.js
```diff
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import axios from 'axios'
```
suzuki: `axios` can remove.

PR #124, opened 2016-09-13
Add word-wrap in comment body

b-kaxa:
### summally
- add `word-wrap: break-word;` in comment body

|before|after|
|---|---|
|||

### notes
- `word-break` has no propety `break-word`
- see [word-break](https://developer.mozilla.org/ja/docs/Web/CSS/word-break) [word-wrap](https://developer.mozilla.org/ja/docs/Web/CSS/word-wrap)


sotarok:
Thank you so much.

PR #126, opened 2016-09-20
Feature mathjax

yktsr:
- Add mathematical notation using [MathJax](https://github.com/mathjax/MathJax)
  - Inline math $ ... $
  - Display math $$ ... $$


sotarok:
Thank you for your p-r!
This would be the great feature!

But, let me consider the way of implementation.
I want to find out another way like followings:

- Implement as a postRenderer of the CrowiRenderer,
- or ... Consider about the structure of Plugins.

Because previews or rendering process is the major problem of the performance.
So let me have time to consider about that.

PR #128, opened 2016-09-25
Modified test utils: Read all models when test will start

suzuki:
## About

- Modifed test utils
- Read all models when test will start


sotarok:
Great!

PR #129, opened 2016-09-25
View all pages for the creator

chuganzy:
## Issue

Non-published pages I created are not listed even in my user page, and it makes it me difficult to find them.

## Solution

List all pages for their creator.


sotarok:
Thank you so much!

PR #142, opened 2016-11-26
Add Node.js version for CircleCI

suzuki:
## About

- Add Node.js version 4.6.2
    - Latest 4.x version

## Motivation

- Crowi is depend on Node.js v4.x
    - But, default version of CircleCI is v0.10 

## References

- http://qiita.com/inuscript/items/d2c87e076595f801b44b


sotarok:
👍

PR #149, opened 2016-12-12
Add "Like" button to smartphone and tablet view

suzuki:
## Overview 

- The "Like" button is not shown in smartphone and tablet view.
    - I would like to "Like" using smartphone.
- No effect in PC view.

### PC

- The "Like" button is not shown in title line.


sotarok:
Thanks!!
I want to consider more about the UI if I could, however, the button appears is better than nothing. 👍

PR #150, opened 2016-12-17
Support the Elasticsearch 5.x

suzuki:
## Overview

- Crowi can not work Elasticsearch 5.x

## Include this pull-request

- Bump up version
    - **v1.6.0**
- Upgrade dependency
    - elasticsearch.js: `11.x` to `12.x`
- Modified old-style mappings
    - `"type" : "multi_field"` is deprecated.
        - https://www.elastic.co/guide/en/elasticsearch/reference/1.4/_multi_fields.html
    - use `string` instead of `multi_field`
        - http://stackoverflow.com/questions/40301061/elasticsearch-5-mapperparserexception-with-multi-field
- Modified query field name
    - use `stored_fields` in Elasticsearch 5.x
        - ~~When ES 2.x, use `fields`. Because ES 2.x does not support `stored_fields`~~
    - I got error messages like this:
```json
{
  "error": {
    "root_cause": [
      {
        "type": "parsing_exception",
        "reason": "The field [fields] is no longer supported, please use [stored_fields] to retrieve stored fields or _source filtering if the field is not stored",
        "line": 1,
        "col": 11
      }
    ],
    "type": "parsing_exception",
    "reason": "The field [fields] is no longer supported, please use [stored_fields] to retrieve stored fields or _source filtering if the field is not stored",
    "line": 1,
    "col": 11
  },
  "status": 400
}
```

## IMPORTANT NOTICE

* Elasticsearch 2.x does **NOT** work on Crowi 1.6.0+

### DIFF
--- a/lib/crowi/index.js
+++ b/lib/crowi/index.js
@@ -344,24 +344,4 @@ Crowi.prototype.exitOnError = function(err) {
   process.exit(1);
 };
 
-Crowi.prototype.getParsedVersion = function(version) {
-  version = version || this.version;
-
-  var major = 0;
-  var minor = 0;
-  var revision = 0;
-
-  if (version.match(/^(\d+)\.(\d+)\.(\d+)$/)) {
-    major    = +(RegExp.$1);
-    minor    = +(RegExp.$2);
-    revision = +(RegExp.$3);
-  }
-
-  return {
-    major: major,
-    minor: minor,
-    revision: revision,
-  };
-};
-
 module.exports = Crowi;
--- a/lib/crowi/index.js
+++ b/lib/crowi/index.js
@@ -217,15 +217,7 @@ Crowi.prototype.getIo = function() {
 };
 
 Crowi.prototype.getSearcher = function() {
-  if (this.searcher !== null) {
-    if (this.searcher.isSupportedESVersion()) {
-      return this.searcher;
-    } else {
-      this.searcher = null;
-    }
-  }
-
-  return null;
+  return this.searcher;
 };
 
 Crowi.prototype.getMailer = function() {
@@ -243,12 +235,21 @@ Crowi.prototype.setupSearcher = function() {
     if (searcherUri) {
       try {
         self.searcher = new (require(path.join(self.libDir, 'util', 'search')))(self, searcherUri);
+        self.searcher
+          .isSupportedESVersion()
+          .then(function(result) {
+            if (result) {
+              resolve();
+            } else {
+              reject(new Error('Detect unsupported Elasticsearch version: ' + self.searcher.getServerVersion()));
+            }
+          });
       } catch (e) {
         debug('Error on setup searcher', e);
         self.searcher = null;
+        resolve();
       }
     }
-    resolve();
   });
 };
 
--- a/lib/crowi/index.js
+++ b/lib/crowi/index.js
@@ -217,7 +217,15 @@ Crowi.prototype.getIo = function() {
 };
 
 Crowi.prototype.getSearcher = function() {
-  return this.searcher;
+  if (this.searcher !== null) {
+    if (this.searcher.isSupportedESVersion()) {
+      return this.searcher;
+    } else {
+      this.searcher = null;
+    }
+  }
+
+  return null;
 };
 
 Crowi.prototype.getMailer = function() {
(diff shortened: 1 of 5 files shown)


sotarok:
To consider this problem, we have to decide the policy which version of es we'll support.

The maintenance const increases if we'll support miltiple versions of es, so IMO, how about the idea below:

- Crowi `1.5.x` supports ES `2.x` (and depends on `11.x`),
- Crowi `1.6.x` supports ES `5.x` (and depends on `12.x`).

suzuki:
It seems like a good idea.

I plan to modify this p-r below:

- Crowi `1.6.x`
    - Check ES version `setupSearcher`
        - https://github.com/crowi/crowi/blob/6ef92d6b1e2c567be3b2fd87e7c6500a5579c55e/lib/crowi/index.js#L227
    - When detect `ES 2.x`
        - Show warning to `console.log()`
        - Set `null` to `self.searcher`
    - Remove `appendFields` method
        - https://github.com/crowi/crowi/pull/150/files#diff-3f4aa453cefa49f6431f1bba3bb53a8eR356

sotarok:
It would be great!

lib/crowi/index.js
```diff
@@ -336,5 +344,24 @@ Crowi.prototype.exitOnError = function(err) {
...
+  version = version || this.version;
+
+  var major = 0;
+  var minor = 0;
+  var revision = 0;
+
+  if (version.match(/^(\d+)\.(\d+)\.(\d+)$/)) {
+    major    = +(RegExp.$1);
+    minor    = +(RegExp.$2);
+    revision = +(RegExp.$3);
```
suzuki: `+( )` means cast to Int

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "crowi",
-  "version": "1.5.2",
+  "version": "1.6.0",
```
suzuki: Bump up version

lib/crowi/index.js
sotarok: In my view, this method is used as both statically and dynamically. (used by es util class with es version number (static), used by the Crowi object with `this.version`)
IMO, the version number parser should not be the member function of Crowi class but be separated as an utility function, and this function just use the function to parse the versioning number `this.version`.

lib/util/search.js
```diff
@@ -19,16 +19,38 @@ function SearchClient(crowi, esUri) {
...
+  var crowiVersion = this.crowi.getParsedVersion();
+  var ESVersion    = this.crowi.getParsedVersion(this.serverInfo.version.number);
+
+  if (crowiVersion.major === 1 && crowiVersion.minor >= 6) {
+    if (ESVersion.major >= 5) {
+      return true;
+    }
+  }
+
+  console.log('WARNING: Detect unsupported Elasticsearch version: ' + this.serverInfo.version.number);
```
sotarok: When unsupported es version detected, Crowi may be good to shutdown.
(Because the operator clearly specified the ES's URI on launching Crowi)

lib/crowi/index.js
suzuki: I created new file:`versionParser.js`.
This method move to that file.

lib/util/search.js
suzuki: I tried to stop execution when detect unsupported version.

suzuki:
## Log

### ES 5.x

- `localhost:9200` is ES 5.x

```
% MONGO_URI=mongodb://localhost/crowi \
  ELASTICSEARCH_URI=localhost:9200 \
  REDIS_URL=localhost \
  FILE_UPLOAD=local \
  DEBUG=crowi:crowi node app.js
[development] Express server listening on port 3000
```

### ES 2.x

- `localhost:9201` is ES 2.4.2

```
% MONGO_URI=mongodb://localhost/crowi \
  ELASTICSEARCH_URI=localhost:9201 \
  REDIS_URL=localhost \
  FILE_UPLOAD=local \
  DEBUG=crowi:crowi node app.js
  crowi:crowi Critical error occured. +0ms
[Error: Detect unsupported Elasticsearch version: 2.4.2]
Error: Detect unsupported Elasticsearch version: 2.4.2
    at /Users/suzuki/work/crowi-develop/crowi/lib/crowi/index.js:244:22
    at process._tickCallback (node.js:368:9)
From previous event:
    at Crowi.setupSearcher (/Users/suzuki/work/crowi-develop/crowi/lib/crowi/index.js:234:10)
    at /Users/suzuki/work/crowi-develop/crowi/lib/crowi/index.js:88:17
    at processImmediate [as _immediateCallback] (timers.js:383:17)
From previous event:
    at Crowi.init (/Users/suzuki/work/crowi-develop/crowi/lib/crowi/index.js:87:6)
    at Object.<anonymous> (/Users/suzuki/work/crowi-develop/crowi/app.js:10:7)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3
```

sotarok:
in #181, support ES 2 and 5, and merged to master by #175
So close this PR thank you

PR #165, opened 2017-01-16
Fix markdown helper bugs in v1.5.3

### LINKED ISSUE #163: [v1.5.3] Completion does not work when entering list symbols
## Overview

- [Firefox] Completion does not work when entering list symbols  (`-`, `*`)
- [Chrome] Can not insert `CR` when cursor on line of list symbols

suzuki:
## Overview

- Fixed #163
- In v1.5.3, some markdown helpers not work.
- List helper ( `-` , `*`)
    - Not work in Firefox, Chrome
- Table helper ( `| foo | bar |` )
    - Not work in Firefox

### DIFF
--- a/resource/js/crowi-form.js
+++ b/resource/js/crowi-form.js
@@ -254,7 +254,10 @@ $(function() {
         }
       }
       //$target.selection('insert', {text: "\n" + listMark, mode: 'before'});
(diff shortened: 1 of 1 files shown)


sotarok:
Thank you, it would be good.

PR #168, opened 2017-01-18
NOT allow to use double slash (`//`) in page path

suzuki:
## Overview

- Now, Crowi not allow to use double slash (`//`) in path.

## Cause

- When Crowi's path has a double slash and nginx's `merge_slashes` is on.
    - User request "double slash path" to nginx.
    - Nginx request "single slash path" to Crowi.
    - Crowi search page by "single slash path" in DB, but not found.

## References

- [merge_slashes](http://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes)
- [nginx はデフォルトで merge slash する](http://r-kurain.hatenablog.com/entry/20120504/1336117840)


sotarok:
Thank you!

PR #171, opened 2017-01-27
Separate blocks with newline characters in revision history

b4b4r07:
Add `newlineIsToken` option to diffLines method

see https://github.com/kpdecker/jsdiff#api

>* `JsDiff.diffLines(oldStr, newStr[, options])` - diffs two blocks of text, comparing line by line.
    Options
>    * `ignoreWhitespace`: `true` to ignore leading and trailing whitespace. This is the same as `diffTrimmedLines`
>    * `newlineIsToken`: `true` to treat newline characters as separate tokens.  This allows for changes to the newline structure to occur independently of the line content and to be treated as such. In general this is the more human friendly form of `diffLines` and `diffLines` is better suited for patches and other computer friendly output.

demo: 

before

after

### DIFF
--- a/resource/js/crowi.js
+++ b/resource/js/crowi.js
@@ -756,7 +756,7 @@ $(function() {
 
           $diffDisplay.text('');
 
-          var diff = jsdiff.diffLines(previousText, currentText);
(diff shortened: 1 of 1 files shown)


sotarok:
Merged by #175

PR #172, opened 2017-01-29
Fixed "Error: Can not find module 'bluebird'"

okonomi:
Fixed that "Error: Can not find module 'bluebird'" occurs at startup when setting file upload locally.


sotarok:
Thank you so much!!

PR #173, opened 2017-01-29
Feature: Github like diff

kaz:
I replaced revision history view with Github-like diff view using [diff2html](https://github.com/rtfpessoa/diff2html).

# Sample

default


sotarok:
Thank you for the great p-r.
I'll check it on review app.

PR #186, opened 2017-02-16
Add pukiwiki like linker

yuki-takei:
Overview
----------

Add the feature to use `[[alias>./relative/path]]` expression in markdown.

Replacement examples
-----------------------

When you write at `/Level1/Level2` page:

```html
<!-- Markdown -->
[[./Level3]]
<!-- HTML -->
<a href="/Level1/Level2/Level3">./Level3</a>


<!-- Markdown -->
[[../AnotherLevel2]]
<!-- HTML -->
<a href="/Level1/AnotherLevel2">../AnotherLevel2</a>


<!-- Markdown -->
Level 3 page is [[here>./Level3]]
<!-- HTML -->
Level 3 page is <a href="/Level1/Level2/Level3">here</a>


<!-- Markdown -->
[[example.com>https://example.com/]]
<!-- HTML -->
<a href="https://example.com/">example.com</a>
```


resource/js/util/PreProcessor/Linker.js
```diff
@@ -1,10 +1,23 @@
...
 
 export default class Linker {
   process(markdown) {
 
     return markdown
       //.replace(/\s(https?:\/\/[\S]+)/g, ' <a href="$1">$1</a>') // リンク
       .replace(/\s<((\/[^>]+?){2,})>/g, ' <a href="$1">$1</a>') // ページ間リンク: <> でかこまれてて / から始まり、 / が2個以上
+      // Pukiwiki Like Linker
+      // see: https://regex101.com/r/k2dwz3/3
+      .replace(/\[\[(([^(\]\])]+)>)?(.+?)\]\]/g, function(all, group1, group2, group3) {
```
sotarok: I prefer `[/xxx/xxx]` (not `[[/xxx]]`)

resource/js/util/PreProcessor/Linker.js
```diff
@@ -1,10 +1,23 @@
...
+        // create url
+        // http(s) から始まっていればそのまま利用、そうでなければ window.location.pathname と連結
+        var url = (group3.match(/^https?:\/\//)) ? group3 : path.join(window.location.pathname, group3);
```
sotarok: It may not work on a list page and a timeline view.
The page would be rendered not only on a single page, but also on any list page. Therefore, join with the current path could make a wrong link.

example:

- if the page is`/foo/bar/baz` and has `./hoge`
    - it should link to `/foo/bar/hoge`
    - it works on a single page
- if user access to `/foo/` (a list page)
    - and show timeline view,
    - then, the link may point to `/foo/hoge`

sotarok: Another thing:
Is `path.join` really needed on a browser? Can't you connect these string simply with '/'?
The script size may be reduced by removing `require('path')`

sotarok:
hmm... let me think.
imo, auto link to internal pages with [ ] is a good idea (imagine that it is used instead of < > ) but the alias syntax is unnessesary.
if people want to make a link with an alias, then markdown style link syntax should be used.
i wouldn'l like to make alternative markdown syntax but comply it

resource/js/util/PreProcessor/Linker.js
yuki-takei: > if the page is/foo/bar/baz and has ./hoge
> 
>    * it should link to /foo/bar/hoge
>    * it works on a single page

First, the motivation of this PR is that we want to use flexible Pukiwiki like path resolution.

https://pukiwiki.osdn.jp/?PukiWiki%E3%81%AE%E4%BD%BF%E3%81%84%E6%96%B9#jee7c1ee

So if the page is /foo/bar/baz,
`[[./hoge]]` generates the link to `/foo/bar/baz/hoge` (not `/foo/bar/hoge`).
This is the intended behavior.

> if user access to /foo/ (a list page)
> 
>   * and show timeline view,
>   * then, the link may point to /foo/hoge

For the above behavior doesn't cause such a problem.

yuki-takei: > Another thing:
> Is `path.join` really needed on a browser? Can't you connect these string simply with '/'?
> The script size may be reduced by removing `require('path')`

**with `require('path')**
```
-rwxrwxrwx 1 yuki yuki 935859  2月 17 18:06 app.39ed906f17808a5a6b6d.js
```

**without `require('path')**
```
-rwxrwxrwx 1 yuki yuki 933848  2月 17 18:11 app.b0a0426e85fb0f2fb9f4.js
```

I think using `path` is bugless way compared to string operation basically.
But if you value the file size, I'll fix.

sotarok:
Finally, I can accept the concept of changing a link syntax from `<>` to `[]`.
However, I don't wanna make an original link syntax like pukiwiki, but want to extend markdown simply.

In particular,

- Can accept a link syntax with `[]` (instead of `<>`, but keep b.c. in near future)
    - then, `[./hoge]` should be a relative link to `hoge` of the same tree,
    - and, `[alias>url]` syntax is not acceptable.

yuki-takei:
I turned the ship toward #189 (Add plugins mechanism).

PR #195, opened 2017-03-09
Omit symlinks

yuki-takei:
My environment
-----------------

* Windows 10 + Windows Subsystem for Linux

What happens?
----------------

* Some assets are not loaded when `node app.js` because of broken symlinks on Windows

Approach
-----------

* Omit symlinks and adopt Environment-independent methods

Note
------

* Omitting `public/js/reveal` has not been done because it looks like beta.

### DIFF
--- a/lib/views/layout/layout.html
+++ b/lib/views/layout/layout.html
@@ -15,6 +15,8 @@
 
   <script src="{{ assets('/js/bundled.js') }}"></script>
   <link href='//fonts.googleapis.com/css?family=Open+Sans:400,700' rel='stylesheet' type='text/css'>
+  <!-- Font Awesome -->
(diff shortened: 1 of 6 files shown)


lib/views/layout/layout.html
```diff
@@ -15,6 +15,8 @@
 
   <script src="{{ assets('/js/bundled.js') }}"></script>
   <link href='//fonts.googleapis.com/css?family=Open+Sans:400,700' rel='stylesheet' type='text/css'>
+  <!-- Font Awesome -->
+  <link href='//cdn.jsdelivr.net/fontawesome/4.7.0/css/font-awesome.min.css' rel='stylesheet' type='text/css'>
```
sotarok: Crowi doesn't want to depend on external resources except googleapis.com, sorry

PR #196, opened 2017-03-12
fix buildServer() called twice

yuki-takei:
see:

* https://github.com/crowi/crowi/blob/master/app.js#L10
* https://github.com/crowi/crowi/blob/master/lib/crowi/index.js#L95
* https://github.com/crowi/crowi/blob/master/lib/crowi/index.js#L301


sotarok:
omg, that is true ... thank you for fix.

PR #197, opened 2017-03-16
Made deleted pages not to be displayed in "recent-create"

b4b4r07:
FEATURE REQUEST:

I often visit "recent-create". Showing deleted pages in this view is annoying. 

I want these pages to be invisible. No problem if merged this patch since it is possible to list the deleted pages from here, I think.

### DIFF
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -541,7 +541,7 @@ module.exports = function(crowi) {
     var User = crowi.model('User');
     var limit = option.limit || 50;
     var offset = option.offset || 0;
-    var conditions = { creator: user._id, redirectTo: null };
+    var conditions = { creator: user._id, redirectTo: null, status: STATUS_PUBLISHED };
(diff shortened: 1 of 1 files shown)


lib/models/page.js
```diff
@@ -541,7 +541,7 @@ module.exports = function(crowi) {
     var User = crowi.model('User');
     var limit = option.limit || 50;
     var offset = option.offset || 0;
-    var conditions = { creator: user._id, redirectTo: null };
+    var conditions = { creator: user._id, redirectTo: null, status: STATUS_PUBLISHED };
```
sotarok: - While the field `status` in Page model has appeared since very recent version, `status` should be specified `null or STATUS_PUBLISHED` to find the page with from old version of crowi.
- Pages that has STATUS_DRAFT or some other statues should be displayed on this page (of course, you know some of these statues are not yet implemented though), so how about `status != deleted` ?

PR #204, opened 2017-04-03
Unfurl Slack URL

mootoh:
To achieve this: https://api.slack.com/docs/message-link-unfurling

### Example

screen shot 2017-04-03 at 12 10 27

### Known issue

- [ ] Cannot unfurl an URL if it contains white space (i.e. `%20`)

### DIFF
--- a/lib/routes/slack.js
+++ b/lib/routes/slack.js
@@ -1,0 +1,95 @@
+module.exports = function(crowi, app) {
+  'use strict';
+
+  var slackHelper = require('../util/slack')(crowi)
+  , Page = crowi.model('Page')
+  , User = crowi.model('User')
+  , actions = {};
+
+  var api = actions.api = {};
+
+  function fetchPage(link) {
+  }
+
+  api.handleEvent = function(req, res) {
+    if (req.body.type != null && req.body.type == 'url_verification') {
+      verifyChallenge(req, res);
+      return;
+    }
+
+    let event = req.body.event;
+    switch (event.type) {
+      case 'link_shared':
+      unfurl(req, res);
+      break;
+      default:
+      break;
+    }
+  }
+
+  function verifyChallenge(req, res) {
+    res.send(req.body.challenge);
+  }
+
+  function unfurl(req, res) {
+    const https = require('https');
+    const HOST = 'slack.com'
+    const PATH = '/api/chat.unfurl';
+
+    const link = req.body.event.links[0];
+    const channel = req.body.event.channel;
+    const url = require('url').parse(link.url, true);
+
+    var pagePath = url.path || null;
+    var pageId = url.query.page_id || null; // TODO: handling
+    var revisionId = url.query.revision_id || null;
+
+    User.findAllUsers()
+    .then (function(users) {
+      return Page.findPage(pagePath, users[0], revisionId);
+    })
+    .then(function(pageData) {
+      let unfurls = { };
+      unfurls[link.url] = {
+        "title": pageData.path,
+        "text": slackHelper.prepareAttachmentTextForCreate(pageData)
+      };
+
+      let config = crowi.getConfig();
+      let token = config.notification['slack:token'];
+      let postData = {
+        "token": token,
+        "channel": channel,
+        "ts": req.body.event.message_ts,
+        "unfurls": encodeURIComponent(JSON.stringify(unfurls))
+      };
+
+      let postDataStr = JSON.stringify(postData);
+
+      let options = {
+        host: HOST,
+        port: 443,
+        path: PATH + "?token=" + postData['token'] + '&channel=' + postData["channel"] + "&ts=" + postData["ts"] + "&unfurls=" + postData["unfurls"],
+        method: 'POST',
+        headers: {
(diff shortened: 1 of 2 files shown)


lib/routes/slack.js
```diff
@@ -0,0 +1,95 @@
...
+
+    var pagePath = url.path || null;
+    var pageId = url.query.page_id || null; // TODO: handling
+    var revisionId = url.query.revision_id || null;
+
+    User.findAllUsers()
+    .then (function(users) {
+      return Page.findPage(pagePath, users[0], revisionId);
+    })
+    .then(function(pageData) {
```
sotarok: You can use `crowi.slack` utility.
IMO, how about create `slack.unfurls()` method onto slack like `slack.post()` method?
like: https://github.com/crowi/crowi/blob/master/lib/util/slack.js#L79

lib/routes/slack.js
```diff
@@ -0,0 +1,95 @@
...
+    const https = require('https');
+    const HOST = 'slack.com'
+    const PATH = '/api/chat.unfurl';
+
+    const link = req.body.event.links[0];
+    const channel = req.body.event.channel;
+    const url = require('url').parse(link.url, true);
+
+    var pagePath = url.path || null;
+    var pageId = url.query.page_id || null; // TODO: handling
```
sotarok: unused var

lib/routes/slack.js
```diff
@@ -0,0 +1,95 @@
...
+    var revisionId = url.query.revision_id || null;
+
+    User.findAllUsers()
+    .then (function(users) {
+      return Page.findPage(pagePath, users[0], revisionId);
```
sotarok: Also, these types of url will be assumed:

- `http://demo.crowi.wiki/58e351c5f06a3b0004e60864`
- `http://demo.crowi.wiki/user/aaaaaaaawrhedb/memo/2017/04/04/dddddddd`

We should handle just page path passed and page id passed as the url.

lib/routes/slack.js
```diff
@@ -0,0 +1,95 @@
...
+        headers: {
+          'Content-Type': 'application/json;charset=utf-8'
+        }
+      };
+
+      let slackReq = https.request(options, (slackRes) => {
+        slackRes.setEncoding('utf8');
+      });
+      slackReq.on('error', (e) => {
+        console.error('failed to send a request to Slack: ' + e.message);
```
sotarok: `debug()` would be preferred

lib/routes/slack.js
sotarok: And also the url that is not the wiki page like `/admin/xxxx` would be passed in case.
In that case, 404 will be returned and no problem, but, a bit it is unexpected behavior.
If we check the path strictly, check `isCreatableName()` before call `findPage()`
https://github.com/crowi/crowi/blob/master/lib/models/page.js#L378

mootoh:
Thanks for the review ❤️  !!! will address those.

btw, how about this? any advice to resolve it?

> Cannot unfurl an URL if it contains white space (i.e. %20)

sotarok:
According to following sample code and its result, it caused by `url.parse()`.

```
var u = require('url');

var urls = [
  'https://demo.crowi.wiki/hoge/Fuga%20Piyo%20XXX',
  'https://demo.crowi.wiki/hoge/memo/2017/%E4%BB%8A%E5%B9%B4%E3%81%AE%E5%8D%A0%E3%81%84%20%E6%9D%A5%E5%B9%B4%E3%81%AE%E5%8D%A0%E3%81%84',
  'https://demo.crowi.wiki/hoge/fuga/piyo-foo-xxx',
];

for (var i = 0, c = urls.length; i < c; i++) {
  var url = decodeURI(urls[i]);
  console.log('Original URL:', url);
  console.log('  Parsed Path:', u.parse(url, true).path);
}
```

```
Original URL: https://demo.crowi.wiki/hoge/Fuga Piyo XXX
  Parsed Path: /hoge/Fuga%20Piyo%20XXX
Original URL: https://demo.crowi.wiki/hoge/memo/2017/今年の占い 来年の占い
  Parsed Path: /hoge/memo/2017/今年の占い%20来年の占い
Original URL: https://demo.crowi.wiki/hoge/fuga/piyo-foo-xxx
  Parsed Path: /hoge/fuga/piyo-foo-xxx
```

I think it is the spec of `url.parse()`
(Furthermore, I also found that this is often discussed. c.f. https://github.com/nodejs/node/issues/2114 )

How about decode `path` with `decodeURI()` after parsed?
Of course this is just a work-around though ...

lib/routes/slack.js
```diff
@@ -0,0 +1,95 @@
...
+  function verifyChallenge(req, res) {
+    res.send(req.body.challenge);
+  }
+
+  function unfurl(req, res) {
+    const https = require('https');
+    const HOST = 'slack.com'
+    const PATH = '/api/chat.unfurl';
+
+    const link = req.body.event.links[0];
```
sotarok: Related to above comment (of about `%20`), is the link passed by Slack already decoded?

PR #219, opened 2017-05-13
Use POSIX-style paths.

Tomasom:
It enables to upload images on Window.

Ref: crowi#217


sotarok:
Thank you for the patch!!
I'll check it on Unix based system and if there are no problem, I'll merge it.

PR #235, opened 2017-06-07
Fixed date format when create new memos

suzuki:
## About

- Changed to use client side date by this PR
    - https://github.com/crowi/crowi/pull/234
- But, date format is incorrect
    - expected: `2017/06/06`

## Screenshot

screenshot 2017-06-06 22 21 47


sotarok:
@suzuki also, can you fix the broken layout?

Original:
image

Broken (from your change):
image

sotarok:
Thank you

PR #237, opened 2017-06-07
Get rid of babel-preset-es2015

abouthiroppy:
Please use `babel-preset-env`. `babel-preset-es2015`  is not recommended.

see: https://github.com/babel/babel-preset-env

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,7 +1,10 @@
 {
   "presets": [
-    "env",
-    "es2015",
+    ["env", {
+      "targets": {
+        "browsers": ["last 2 versions"]
+      }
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,7 +1,10 @@
 {
   "presets": [
-    "env",
-    "es2015",
+    ["env", {
+      "targets": {
+        "browsers": ["last 2 versions"]
```
abouthiroppy: @sotarok What do you think about the support version?

sotarok: I think, it whould be ok! > last 2 versions

PR #246, opened 2017-07-19
Prepare for 1.6.3

sotarok:


abouthiroppy:
I think we should use conventional-changelog because CHANGELOG should be generated automatically;)

https://github.com/conventional-changelog/conventional-changelog

@sotarok What do you think?

sotarok:
@abouthiroppy  That looks nice.
Should we comply the commit comment if we use `conventional-changelog`?

abouthiroppy:
Yeah, we should set a rule for `conventional-changelog`.
I recommend an angular style because many projects adopt it.
FYI: https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit

PR #254, opened 2017-08-21
[WIP] Feature/make use of showdown

hiroppy:
We can use GitHub's markdown syntax now!!!

see: https://github.com/showdownjs/showdown


abouthiroppy:
new options: https://github.com/showdownjs/showdown#valid-options

resource/js/util/CrowiRenderer.js
```diff
@@ -96,35 +96,12 @@ export default class CrowiRenderer {
   parseMarkdown(markdown, dom) {
     let parsed = '';
 
-    const markedRenderer = new marked.Renderer();
-    markedRenderer.code = this.codeRenderer;
```
sotarok: We have to think about syntax highlight.

abouthiroppy: Good catch.

abouthiroppy: I'll add highlight.js as an extension;)

sotarok:
We're thinking how about move forward to remark https://github.com/wooorm/remark

abouthiroppy:
OK, I'll try it.

sotarok:
We'll plan to use remark.js as next markdown parser so close it.
Thank you!

PR #256, opened 2017-08-22
feat(api): Api pages.get handles page_id

sotarok:
- `pages.get` now handles page_id parameter

### DIFF
--- a/lib/routes/page.js
+++ b/lib/routes/page.js
@@ -652,12 +652,22 @@ module.exports = function(crowi, app) {
    * @apiParam {String} revision_id
    */
   api.get = function(req, res){
-    var pagePath = req.query.path || null;
-    var pageId = req.query.page_id || null; // TODO: handling
-    var revisionId = req.query.revision_id || null;
+    const pagePath = req.query.path || null;
+    const pageId = req.query.page_id || null; // TODO: handling
+    const revisionId = req.query.revision_id || null;
 
-    Page.findPage(pagePath, req.user, revisionId)
-    .then(function(pageData) {
+    if (!pageId && !pagePath) {
+      return res.json(ApiResponse.error(new Error('Parameter path or page_id is required.')));
+    }
+
+    let pageFinder;
+    if (pageId) { // prioritized
+      pageFinder = Page.findPageByIdAndGrantedUser(pageId, req.user);
+    } else if (pagePath) {
+      pageFinder = Page.findPage(pagePath, req.user, revisionId);
+    }
+
+    pageFinder.then(function(pageData) {
       var result = {};
       result.page = pageData;
 
--- a/test/models/page.test.js
+++ b/test/models/page.test.js
@@ -6,22 +6,22 @@ var chai = require('chai')
   ;
 chai.use(sinonChai);
 
-describe('Page', function () {
+describe('Page', () => {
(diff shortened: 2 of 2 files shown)


test/models/page.test.js
```diff
@@ -279,4 +283,40 @@ describe('Page', function () {
...
+          expect(pageData.path).to.equal(pageToFind.path);
+          done();
+        });
+      });
+
+      it('should error by grant', done => {
+        const pageToFind = createdPages[0];
+        const grantedUser = createdUsers[1];
+        Page.findPageByIdAndGrantedUser(pageToFind._id, grantedUser)
+        .then(pageData => {
```
abouthiroppy: This line is unnecessary.

test/models/page.test.js
```diff
@@ -279,4 +283,40 @@ describe('Page', function () {
     });
   });
 
+  describe('.findPage', function () {
```
abouthiroppy: nit: Arrow function

test/models/page.test.js
```diff
@@ -279,4 +283,40 @@ describe('Page', function () {
...
+    context('findPageById', () => {
+      it('should find page', function(done) {
```
abouthiroppy: ditto

test/models/page.test.js
```diff
@@ -279,4 +283,40 @@ describe('Page', function () {
...
+        const pageToFind = createdPages[0];
+        Page.findPageById(pageToFind._id)
+        .then(pageData => {
+          expect(pageData.path).to.equal(pageToFind.path);
+          done();
+        });
+      });
+    });
+
+    context('findPageByIdAndGrantedUser', function() {
```
abouthiroppy: ditto

test/models/page.test.js
```diff
@@ -279,4 +283,40 @@ describe('Page', function () {
...
+      it('should find page', function(done) {
```
abouthiroppy: ditto

lib/routes/page.js
```diff
@@ -652,12 +652,22 @@ module.exports = function(crowi, app) {
...
-    var pagePath = req.query.path || null;
-    var pageId = req.query.page_id || null; // TODO: handling
-    var revisionId = req.query.revision_id || null;
+    const pagePath = req.query.path || null;
+    const pageId = req.query.page_id || null; // TODO: handling
+    const revisionId = req.query.revision_id || null;
 
-    Page.findPage(pagePath, req.user, revisionId)
-    .then(function(pageData) {
+    if (!pageId && !pagePath) {
```
abouthiroppy: Question: Is there a possibility that `pageId` is 0 and `pagePath` is 0?

sotarok: Yes, it is possible beacuse this is an API that is called by any client.
If the client doesn't specify both `page_id` and `path`, `req.query` doesn't have enough parameter, so we need to validate them.

test/models/page.test.js
sotarok: Added done with error below. 
(Because the test should be failed if the user found)

lib/routes/page.js
abouthiroppy: Will you add verification at this PR?

sotarok: Yes, I will

PR #257, opened 2017-09-02
Added confirmation of existence of email when updating user info

suzuki:
## About

- fix #250

PR #259, opened 2017-09-19
Fix a problem that could not update without changing email

suzuki:
## About

- I injected a problem by #257.
- It is impossible update self information without changing email.

## How to fix the problem

- Added self `user._id` into query to find same email address

### DIFF
--- a/lib/routes/me.js
+++ b/lib/routes/me.js
@@ -89,9 +89,9 @@ module.exports = function(crowi, app) {
         return res.render('me/index', {});
       }
 
-      User.findOne({email: email, _id: {'$ne': userData._id}}, (err, existingUserData) => {
+      User.findOne({email: email}, (err, existingUserData) => {
         // If another user uses the same email, an error will occur.
-        if (existingUserData) {
+        if (existingUserData && !existingUserData._id.equals(userData._id)) {
           debug('Email address was duplicated');
           req.form.errors.push('It can not be changed to that mail address');
           return res.render('me/index', {});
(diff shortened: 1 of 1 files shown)


lib/routes/me.js
```diff
@@ -89,7 +89,8 @@ module.exports = function(crowi, app) {
         return res.render('me/index', {});
       }
 
-      User.findOne({email: email}, (err, existingUserData) => {
+      User.findOne({email: email, _id: {'$ne': userData._id}}, (err, existingUserData) => {
```
sotarok: hmm ... ?
if the user couldn't find here, then the process is not into this if statement, 
it means ... lang and name are also not updated?

sotarok: it should be ...

- user changes their email address to another one, and it's not duplicated => email can be changed
- user changes email address and it's duplicated => email can't be updated
- anyway, others (name and lang) can be updated

sotarok:
@suzuki it has also bug. can you review it?

lib/routes/me.js
abouthiroppy: First of all, we should write exception handling, right?😉

suzuki: when email was invalid, `lang` and `name` are not updated.
https://github.com/crowi/crowi/pull/259/files#diff-1972c53f45200d68ac6d5e645409aca1L87

here is same logic.

sotarok: I misunderstood about this logic.
It would be ok.

PR #273, opened 2018-01-11
Clear google session if user data is not found

chuganzy:
We have no choice but to clear the cookie when accidentally selected another account to login with in Google login flow so far which is not user-friendly.
To avoid it, this PR makes it to clear google session if user data is not found.


sotarok:
Thank you!

PR #285, opened 2018-05-11
Development environment with docker

lightnet328:
# Overview
Getting rid of environmental difference and improving development efficiency, I think we should use Docker.

`docker-compose.development.yml` is optimized for development, the server restarts and assets(Style, Script) is compiled when a file changes.

## Usage
```
$ docker-compose -f docker-compose.development.yml up
```


elasticsearch/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch:6.2.2
```
sotarok: Please set version 5 for Crowi 1.6.x

PR #286, opened 2018-05-11
Fix a problem that we can't change a page to portal in a specific condition

### LINKED ISSUE #283: Can not unportalize
## About

- I can not unportalize my potal page.

## Reproduce

1. Open a portal page

screenshot 2018-03-27 09 38 28

2. Click the spanner icon

screenshot 2018-03-27 09 38 32

3. The modal opened and click the "Unportalize!" button

screenshot 2018-03-27 09 38 35

4. The alert appeared.
screenshot 2018-03-27 09 38 39

lightnet328:
# Overview
Fix a problem that we can't change a page to portal in a specific condition.

# Reproduce
1. Create a page (`/hoge`)

2. Move this page to `/hoge/`

* A page whose URL ends with a slash will be a portal

3. We can't change a page to portal

# Cause
`/hoge` page become a redirect page when moved this page to `/hoge/`.
Page is exists as a redirect page, so we can't unportalize `/hoge/`.

# Solution (Changes)
- If existed page is a redirect page, we'll delete it.
  - By this, "Unportalize" will work normally.
-  In case of renaming, if new page's url ends with slash, the server don't create a redirect page.

# Related issues
Close #283

### DIFF
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -138,7 +138,7 @@ module.exports = function(crowi) {
   };
 
   pageSchema.methods.isRedirectOriginPage = function() {
-    return this.redirectTo != null;
+    return this.redirectTo !== null;
   };
 
   pageSchema.methods.like = function(userData) {
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -137,6 +137,10 @@ module.exports = function(crowi) {
     });
   };
 
+  pageSchema.methods.isRedirectOriginPage = function() {
+    return this.redirectTo != null;
+  };
+
   pageSchema.methods.like = function(userData) {
     var self = this,
       Page = self;
--- a/lib/routes/page.js
+++ b/lib/routes/page.js
@@ -860,8 +860,9 @@ module.exports = function(crowi, app) {
     var pageId = req.body.page_id;
     var previousRevision = req.body.revision_id || null;
     var newPagePath = Page.normalizePath(req.body.new_path);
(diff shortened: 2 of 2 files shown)


lib/models/page.js
```diff
@@ -137,6 +137,10 @@ module.exports = function(crowi) {
     });
   };
 
+  pageSchema.methods.isRedirectOriginPage = function() {
+    return this.redirectTo != null;
```
hiroppy: Please use `!== null` if `this.redirectTo` has only `null`.

lib/routes/page.js
```diff
@@ -892,6 +888,26 @@ module.exports = function(crowi, app) {
...
+    Page.findPageByPath(newPagePath)
+    .then(function(page) {
+      if (page.isRedirectOriginPage() &&  page.isGrantedFor(req.user)) {
+        debug('Unlink page', page._id, page.path);
+        Page.removePageById(page._id)
+        .then(function(data) {
+          debug('Redirect Page deleted', data.path);
+          rename();
+        }).catch(function(err) {
+          debug('Error occured while get setting', err, err.stack);
```
sotarok: Is this debug message correct? 🤔

sotarok:
Oops, sorry please fix the code commented by @hiroppy :)

lib/models/page.js
lightnet328: I am too embarrassed.
Thank you for pointing it out! I fixed :)

PR #294, opened 2018-06-04
Internationalization frontend

lightnet328:
# Overview
Supporting to several languages in React component, use i18next same as backend.

# Reproductions
1. Import `react-i18next`
```js
import { translate } from 'react-i18next';
```

2. Export a component wrapped with `translate` HoC
```js
export default translate()(PageHistory);
```

3. Translate a message by `t` prop
```js
{this.props.t("Move")}
```

# Screens


sotarok:
Super rally nice, I love it ❤️

PR #299, opened 2018-06-10
Support bearer token

hirakiuc:
This pull req just add a `Bearer token support in Authorization header` to crowi.

NOTE: [the format of Authorization header with Bearer token](https://tools.ietf.org/html/rfc6750#section-2.1)

### DIFF
--- a/lib/util/middlewares.js
+++ b/lib/util/middlewares.js
@@ -187,10 +187,30 @@ exports.loginRequired = function(crowi, app) {
   };
 };
 
+/**
+ * Extract Bearer token from Authorization header.
+ *
+ * @param {Object} headers HTTP request headers
+ * @param {string} headers.authorization Authorization header value.
+ * @return {?string} found access_token or null.
+ */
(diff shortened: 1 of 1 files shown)


lib/util/middlewares.js
```diff
@@ -187,10 +187,30 @@ exports.loginRequired = function(crowi, app) {
...
 
+/**
+ * Extract Bearer token from Authorization header.
+ *
+ * @param {Object} headers HTTP request headers
+ * @param {string} headers.authorization Authorization header value.
+ * @return {?string} found access_token or null.
+ */
+var extractBearerToken = (headers) => {
+  const v = headers.authorization;
```
sotarok: Comment: here useing `const`, so `var` of `var extractBearerToken` also should be `const`?

hirakiuc:
Oh, sorry.
I changed that 'var' with `const`. 👍

sotarok:
Thank you for the great PR. Merged!

PR #301, opened 2018-06-11
Rename the page hierarchy

lightnet328:
# Overview
Implement the feature to rename the page hierarchy.
ex. `/products/sushi/*` -> `/products/pizza/*`
The feature seems simple at first glance, but in fact, it is a bit complicated, so I will explain the specification while showing screens.

# Screens
localhost_3000_products_
User can open modal from Page, Portal and PageList.

localhost_3000_products_ 1
A list of pages to be changed is displayed when a user enters a new path

localhost_3000_products_ 2
If `/products/APL-ZS/design` exists, it becomes the parent of `/products/APL-ZS/desing/*`

localhost_3000_products_ 3
If there are pages that can't be renamed, the reason is displayed


lib/models/page.js
```diff
@@ -183,6 +183,22 @@ module.exports = function(crowi) {
     })
   }
 
+  // Unlink: Remove redirect origin page
+  pageSchema.methods.unlink = async function(userData) {
+    const Page = crowi.model('Page')
+    if (this.isUnlinkable(userData)) {
+      debug('Unlink page', this._id, this.path)
+      try {
+        const redirectPage = await Page.removePageById(page._id)
```
sotarok: Here 

> ReferenceError: page is not defined

occurred.
It happened when rename the page to the path which exists as the redirect page.

lib/routes/page.js
```diff
@@ -892,65 +892,156 @@ module.exports = function(crowi, app) {
...
+    return paths.map(({ path }) => [path, renamePath(path)]).reduce((l, [k, v]) => Object.assign(l, { [k]: v }), {})
+  }
+
+  async function checkPagesRenamable(pathMap, user) {
+    let error = false
+    let errors = {}
+    for (let [oldPath, newPath] of Object.entries(pathMap)) {
+      let e = []
+      if (!Page.isCreatableName(newPath)) {
+        e.push('作成できないページ名です')
```
sotarok: This should be translated.
(Also Japanese messages below)

lightnet328:
I added some tests and fixed bugs for the backend.
- It is possible to rename a tree without limited to a number of pages
- Even if a page started with the name of another page, it no longer treats it as a child
  - e.g. `/car` and `/carrot`

lib/views/page.html
```diff
@@ -142,6 +142,9 @@ <h1 class="title flex-item-title">{{ path|insertSpaceToEachSlashes }}</h1>
       </a>
       <ul class="dropdown-menu">
        <li><a href="#" data-target="#renamePage" data-toggle="modal"><i class="fa fa-share"></i> {{ t('Move') }}</a></li>
+       {% if !isTopPage() and !isUserPage(path) %}
+       <li><a href="#" data-target="#renameTree" data-toggle="modal"><i class="fa fa-share"></i> {{ t('Move Tree') }}</a></li>
```
sotarok: How about, `Move pages under /hogehogehoge` ?
User can't recognize pages list as `Tree`

lib/routes/page.js
```diff
@@ -889,65 +889,96 @@ module.exports = function(crowi, app) {
...
-            throw new Error("Someone could update this page, so couldn't delete.")
-          }
+    const rename = async function() {
+      try {
+        const page = await Page.findPageById(pageId)
+        if (!page.isUpdatable(previousRevision)) {
+          return res.json(ApiResponse.error(new Error("Someone could update this page, so couldn't delete.")))
+        }
+        const result = {
+          page: await Page.rename(page, newPagePath, req.user, options),
```
sotarok: This change of `result.page` causes an error on rename page action.

Originally, `result.page` was page object itsef,
https://github.com/crowi/crowi/pull/301/files#diff-db7e2375060f4b1d51ec816b5d6035bbL919

but this result of `await Page.rename` is the result of update of mongodb.

lightnet328: I reverted this change 🙇

PR #303, opened 2018-06-15
Use docker image again

lightnet328:
# Overview
Since it is necessary to unify the version of npm, I use Crowi's Docker image again.

# Historical Background
Problems occurred when switching branches by using Named Volume in the compose file. Therefore, we decided to use the Node image.
https://github.com/crowi/crowi/pull/291


sotarok:
It works :)

PR #310, opened 2018-07-02
Use the language detected from the server

lightnet328:
# Problem
Language rendered on client side and server side may be different.
Because both are detecting languages based on different information.

# Solution
The client side uses the language detected by the server side.

### DIFF
--- a/client/i18n.js
+++ b/client/i18n.js
@@ -1,26 +1,39 @@
 import i18n from 'i18next'
-import LanguageDetector from 'i18next-browser-languagedetector'
+import LngDetector from 'i18next-browser-languagedetector'
 import { reactI18nextModule } from 'react-i18next'
 import en from 'locales/en/translation.json'
 import ja from 'locales/ja/translation.json'
 
-i18n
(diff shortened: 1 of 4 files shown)


client/i18n.js
```diff
@@ -1,26 +1,39 @@
...
+    cacheUserLanguage(lng, options) {},
   })
 
-export default i18n
+  i18n
+    .use(lngDetector)
+    .use(reactI18nextModule)
+    .init({
+      fallbackLng: 'en',
+      debug: true,
```
dyong0: [IMO] I believe this is out of scope of the PR. But it'd be nice to hide debug logs in production.
```js
debug:  process.env.NODE_ENV && (process.env.NODE_ENV === 'development')
```
https://github.com/crowi/crowi/blob/master/tools/webpack.client.js#L81-L85

lightnet328: `process.env.NODE_ENV` is defined only in the production environment.
But since it does not seem to be used everywhere, I'll changes it with this pull request.

PR #312, opened 2018-07-13
Group search results by page type

lightnet328:
# Overview
When the number of pages increases, it becomes difficult to find the target page by searching.
In order to simplify the search, I grouped the search results by following page types.

# Page Types
- User
    - Page path starts with `/user/`
- Portal
    - Page path ends with `/`
    - NOT User page
- Public
    - Page path does not end with `/`
    - NOT User page

# Screentshots

## Suggests
localhost_3000_ ipad

## Results
localhost_3000__search_q e9 96 8b e7 99 ba type portal ipad


lib/util/search.js
```diff
@@ -344,17 +364,28 @@ SearchClient.prototype.appendResultSize = function(query, from, size) {
   query.size = size || this.DEFAULT_LIMIT
 }
 
-SearchClient.prototype.appendCriteriaForKeywordContains = function(query, keyword) {
+SearchClient.prototype.initializeBoolQuery = function(query) {
```
sotarok: 👍

client/components/SearchPage/SearchToolbar.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'react-i18next'
+import { Nav, NavItem } from 'react-bootstrap'
+import Icon from 'components/Common/Icon'
+
+class SearchToolbar extends React.Component {
+  constructor() {
+    super()
```
sotarok: ❓  `SearchToolbar` doesn't catch `props` on constructor, is it okay?

Upper level component passes props though ...
https://github.com/crowi/crowi/pull/312/files#diff-24c2fd48e6e519712edba1bff9de226bR95

sotarok: I fixed this on 53e4c0a

PR #313, opened 2018-07-13
Share some pages externally

lightnet328:
# Overview
Share some pages externally.
This feature aims to share specific pages for users who do not have an account.

## Notice
Therefore, This is not a feature to allow access to all pages without an account.
Ref: https://github.com/crowi/crowi/issues/155

### Share page

1. Users create shared links from pages they want to share.
2. If necessary, users can set keywords necessary for viewing the page without an account.
3. Users tell a link to a person.

### View page

1. People who receive the link access that link.
    - If a secret keyword is set, the user inputs it
    - If it's not set, people can view the page

## Screenshots
localhost_3000_hr_joiningprocedures ipad

PR #314, opened 2018-07-20
Upgrade to webpack v4

lightnet328:
# Problem
Since UglifyJS included in the webpack currently used does not correspond to the syntax of ES6, the production build may fail.

# Solution

First of all, we need to use the new version of UglifyJS in order to prevent the production build from failing using ES6 syntax. And there are ways to use the new version of UglifyJS as follows.

1. Use an independent package
    - [webpack-contrib/uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin)
2. Upgrade to webpack v4
    - **This Pull Request proposes this method**
    - webpack v4 contains a new UglifyJS
3. Use an independent package and later use the packages contained in webpack


package.json
```diff
@@ -127,27 +125,29 @@
...
     "dev:server": "DEBUG=crowi:* NODE_ENV=development nodemon --watch ./lib/ --watch ./locales/ --ext 'js,json' --exec node app.js",
-    "dev:client": "NODE_ENV=development webpack --progress --profile --colors --display-error-details --watch --config tools/webpack.config.js",
+    "dev:client": "NODE_ENV=development webpack --progress --profile --colors --display-error-details --watch --mode=development --config tools/webpack.config.js",
     "flow": "flow",
     "format": "eslint --fix \"**/*.js\"",
     "lint": "eslint \"**/*.js\"",
     "test": "mocha -r test/bootstrap.js --globals chai --reporter dot test/**/*.test.js",
     "build": "npm run prod:client",
-    "prod:client": "NODE_ENV=production webpack --bail --progress --colors --display-error-details --config tools/webpack.config.js",
+    "prod:client": "NODE_ENV=production webpack --bail --progress --colors --display-error-details --mode=production --config tools/webpack.config.js",
```
sotarok: I removed `--progress` on #313 :)

package.json
```diff
@@ -127,27 +125,29 @@
...
+    "sinon-chai": "~2.12.0",
+    "webpack-cli": "^3.0.8"
   },
   "license": "MIT",
   "scripts": {
     "start": "node app.js",
     "dev": "concurrently --kill-others \"npm run dev:server\" \"npm run dev:client\" \"npm run watch\" ",
     "dev:server": "DEBUG=crowi:* NODE_ENV=development nodemon --watch ./lib/ --watch ./locales/ --ext 'js,json' --exec node app.js",
-    "dev:client": "NODE_ENV=development webpack --progress --profile --colors --display-error-details --watch --config tools/webpack.config.js",
+    "dev:client": "NODE_ENV=development webpack --progress --profile --colors --display-error-details --watch --mode=development --config tools/webpack.config.js",
```
dyong0: DRY, mode is the same as `NODE_ENV`
```js
const env = process.env.NODE_ENV

// webpack config
{
    mode: env
}
```

lightnet328: Surely. I replaced it with environment variables.

package.json
lightnet328: Thanks for letting me know.
Since `mode` option refers to an environment variable, merge conflict does not occur now.

PR #322, opened 2018-08-09
update readme from node6.x to 8.x

inductor:
got this error when trying installation with Node.js 6.9.5
```
error crowi@1.7.0-dev: The engine "node" is incompatible with this module. Expected version "8.x".
error Found incompatible module
```


sotarok:
Thank you!

PR #349, opened 2018-09-21
Improve login flow

lightnet328:
# Features
## Restrict login method
- Force third-party authentication settings
- Prohibit password authentication

\* Those can be enabled or disabled from the admin page

## Change login flow
- Only social authentication buttons are displayed on the login page when password authentication is prohibited
  - If the login is successful but not associated with an account, a user will be redirected to the registration page

## Validate access tokens in periodically
- Log out if access token refreshing is failed

## Sync login status among multiple tabs
- https://github.com/crowi/crowi/pull/338


lib/routes/login.js
```diff
@@ -99,11 +104,18 @@ module.exports = function(crowi, app) {
         }
       })
     } else {
+      const { user = {} } = req
+      const User = crowi.model('User')
+      const isLoggedIn = user && '_id' in user && user.status === User.STATUS_ACTIVE
```
sotarok: Comment: good to be a util function

lib/routes/me.js
```diff
@@ -204,16 +200,28 @@ module.exports = function(crowi, app) {
...
+  }
+
   actions.authGoogle = function(req, res) {
     const googleAuth = require('../util/googleAuth')(config)
     const { user: userData } = req
     const toDisconnect = !!req.body.disconnectGoogle
     const toConnect = !!req.body.connectGoogle
+    const callback = req.session.callback || '/me'
 
     if (toDisconnect) {
```
sotarok: Is it okay to disconnect from Google when `disablePasswordAuth` set `true`?
(GitHub auth as well)
Check the setting and validate here to prevent disconnect and be impossible to login.

lib/util/auth.js
```diff
@@ -0,0 +1,34 @@
+module.exports = {
+  isAccessTokenExpired(req) {
+    const { auth = {} } = req.session
+    const { expiryDate = null } = auth
+    if (expiryDate === null) {
+      return false
```
sotarok: ❓  is `false` okay if `expiryDate` is null ?

sotarok: memo:

When the password auth and the third-party auth enabled at the same time and third party auth expired, should it jump to login page even if there is remaining password auth session?

lib/util/middlewares.js
```diff
@@ -160,10 +160,29 @@ exports.adminRequired = function() {
...
-    var User = crowi.model('User')
+  return async function(req, res, next) {
+    const User = crowi.model('User')
+    const config = crowi.getConfig()
+    const { path = '', originalUrl } = req
+    const auth = require('./auth')
+    const query = originalUrl === '/' ? '' : `?continue=${originalUrl}`
+    const isAuthPage = path.startsWith('/me/auth/')
+
+    if (!isAuthPage && auth.isAccessTokenExpired(req)) {
```
sotarok: memo: confirm the behavior of password auth

lib/util/middlewares.js
```diff
@@ -177,13 +196,11 @@ exports.loginRequired = function(crowi, app) {
     }
 
     // is api path
-    var path = req.path || ''
     if (path.match(/^\/_api\/.+$/)) {
```
sotarok: here we can also use `isAPI` :)

lib/util/url.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const getContinueUrl = (req = {}) => {
+  const { body = {}, query = {} } = req
+  const url = body.continue || query.continue
+  const continueUrl = /^(?![/\\]{2,})\/.*$/.test(url) ? url : '/'
+  return continueUrl
+}
+
+module.exports = { getContinueUrl }
```
sotarok: 💬 テスト欲しい感があります

PR #362, opened 2018-10-03
fix: strict startWith check

otofune:
Can't show moved tree, when target tree has document with conditions following:

- let P is parent path.
- let SR is list of same rank pathes of P.
- There are pathes staring with P in SR.


lightnet328:
I confirmed the reproduction of this problem and its correction.
LGTM

PR #374, opened 2018-11-01
Import blockquote styles from bootstrap3

lightnet328:
# Overview
Bootstrap 4 does not have the style of blockquote.

# Screens

## Before

blockquote-before

## After 

blockquote-after


lightnet328:
Thanks for reviewing. I added a reference.

PR #384, opened 2018-11-27
Improve page performance

sotarok:
- Remove unused code
- Separate route function for page/user_page.

### DIFF
--- a/lib/views/user_page.html
+++ b/lib/views/user_page.html
@@ -74,6 +74,11 @@
 </div>
 
 {% else %}
+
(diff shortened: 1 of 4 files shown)


lib/views/user_page.html
```diff
@@ -74,6 +74,11 @@ <h2>{{ pageUser.name }}</h2>
 </div>
 
 {% else %}
+
+  <div class="alert alert-danger">
```
otofune: 👍

PR #387, opened 2018-11-29
Select a mapping file automatically when build index

lightnet328:
# Overview
Select a mapping file automatically to support more various elasticsearch environments.

## How to select

- "analysis-kuromoji" is installed
    - Use "mapping-kuromoji.json"
- "analysis-sudachi" is installed
    - Use "mapping-kuromoji.json"
- the others
    - Use "mapping.json"

"mapping.json" is renamed to "mapping-kuromoji.json".
Instead, a mapping file no using "analysis-kuromoji"  is created as "mapping.json".


lib/util/search.js
```diff
@@ -49,8 +50,16 @@ function SearchClient(crowi, esUri) {
   })
 
   this.registerUpdateEvent()
+}
 
-  this.mappingFile = crowi.resourceDir + 'search/mappings.json'
+SearchClient.prototype.selectMappingFile = function() {
```
sotarok: This function is depending on the result of `checkESVersion()` that is called in the bootstrap process of `lib/crowi/index.js`.
Is it needed to check wether the plugin information loaded correctly or not?

lightnet328: `selectMappingFile()` return 'mappings.json' in default.
So it may not be necessary to know whether `checkESVersion()` ran in the past or not.
I think run `checkESVersion()` before `selectMappingFile()` is more kind (can get the latest information).

PR #391, opened 2018-12-03
Improve the calculation method of search score

lightnet328:
# Overview

## Change the calculation method of search score

A score based on the bookmark number is added to the conventional score.

## Improvement of UI

(Implemented by @sotarok)

### Building index progress bar
localhost_3000_admin_search

### DIFF
--- a/lib/models/bookmark.js
+++ b/lib/models/bookmark.js
@@ -1,10 +1,9 @@
 module.exports = function(crowi) {
-  var debug = require('debug')('crowi:models:bookmark')
-  var mongoose = require('mongoose')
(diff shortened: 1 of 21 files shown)


lib/models/bookmark.js
```diff
@@ -147,58 +143,49 @@ module.exports = function(crowi) {
...
-          }
-          return reject(err)
-        }
+  bookmarkSchema.statics.add = async function(page, user) {
+    const Bookmark = this
 
-        resolve(bookmark)
-      })
-    })
+    const newBookmark = new Bookmark({ page, user, createdAt: Date.now() })
```
sotarok: 👍

PR #393, opened 2018-12-06
Fix an issue that error occurs in admin/share page after a shared page was completely deleted

lightnet328:
# Overview
Page ref becomes null when populate share document after shared page deleted.
Therefore, an error occurs in admin/share page.
In this PR, fix this issue.

# Changes

- Show completely deleted pages path as '(Deleted)'

# Screens

localhost_3000_admin_share ipad

### DIFF
--- a/lib/models/page.js
+++ b/lib/models/page.js
@@ -928,36 +928,25 @@ module.exports = function(crowi) {
   /**
    * This is danger.
    */
-  pageSchema.statics.completelyDeletePage = function(pageData, user, options) {
+  pageSchema.statics.completelyDeletePage = async function(pageData, user, options) {
     // Delete Bookmarks, Attachments, Revisions, Pages and emit delete
-    var Bookmark = crowi.model('Bookmark')
-    var Attachment = crowi.model('Attachment')
-    var Comment = crowi.model('Comment')
-    var Page = this
-    var pageId = pageData._id
+    const Bookmark = crowi.model('Bookmark')
+    const Attachment = crowi.model('Attachment')
+    const Comment = crowi.model('Comment')
+    const Page = this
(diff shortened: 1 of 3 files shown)


lib/models/page.js
```diff
@@ -928,36 +928,25 @@ module.exports = function(crowi) {
...
-          pageEvent.emit('delete', pageData, user) // update as renamed page
-          resolve(pageData)
-        })
-        .catch(reject)
-    })
+    await Bookmark.removeBookmarksByPageId(pageId)
+    await Attachment.removeAttachmentsByPageId(pageId)
+    await Comment.removeCommentsByPageId(pageId)
+    await Page.removePageById(pageId)
+    await Page.removeRedirectOriginPageByPath(pageData.path)
```
sotarok: No need to wrap with try-catch?

lightnet328: I think followings are all behave in the same.

```js
return new Promise((resolve, reject) => {
    AsyncFunc()
    .then(() => AsyncFunc())
    // ...
    .catch(reject)
})
```

```js
try {
    await AsyncFunc()
    await AsyncFunc()
    // ...
} catch (err) {
    throw err
}
```

```js
await AsyncFunc()
await AsyncFunc()
// ...
```

We probably need to `try-catch` if transactions and logging are necessary.

otofune: I think It seems unnecessary to catch, [the old code did the same](https://github.com/crowi/crowi/pull/393/files#diff-79d809949dd96cac2e1cbbeecdf7c6ecL959) 👀
But it may be good to decide the policy now.

sotarok: Ah, ok! I misunderstood.

PR #398, opened 2018-12-14
Update to mongoose@5 and call Model#syncIndexes instead of autoIndex

otofune:
To prevent duplication of indexes.

`Model#syncIndexes` can drop indexes that not listed in schemas.
This feature is added at v5.2. For more detail, see [CHANGELOG](https://github.com/Automattic/mongoose/blob/master/History.md#520--2018-07-02).


otofune:
I saw the [migration to v5 document](https://github.com/Automattic/mongoose/blob/master/migrating_to_5.md#return-values-for-remove-and-deletex), there are no inconvenient I think.
But I'll check this PR with dump from working crowi server, please look forward to the good news.

package-lock.json
```diff
@@ -4373,7 +4373,7 @@
         },
         "string_decoder": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
+          "resolved": "http://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
```
Sw0rdstream: change to http will incur the risk of man-in-middle attack

otofune: Thank you! This problem happened by npm registry API. I will change `http://` to `https://`.

otofune:
This PR works fine on env created after https://github.com/crowi/crowi/commit/c7b78c418d06364de40704a9943b713ad5dae1fe (before this commit, multiple pages can have same path...)
I guess this PR can be merged into v1.8, but we must create migration files before v1.8 become mainline.

sotarok:
It would be nice.
Could you also please update README and dependencies?

PR #401, opened 2018-12-20
Introduce tj/migrate

otofune:
- [x] Add tj/migrate support
  * [x] Support MongoDB as persistent backend
- [x] Add migration for changes at https://github.com/crowi/crowi/commit/c7b78c418d06364de40704a9943b713ad5dae1fe

### DIFF
--- a/migrate-mongo-store.js
+++ b/migrate-mongo-store.js
@@ -40,17 +40,14 @@ const migrationSchema = new mongoose.Schema({
   ],
 })
 migrationSchema.methods.toEntry = function() {
-  const m = this
-  const entry = {
-    title: m.title,
-    timestamp: m.status === 'up' ? m.at.getTime() : null,
+  return {
+    title: this.title,
+    timestamp: this.status === 'up' ? this.at.getTime() : null,
   }
-  return entry
 }
 migrationSchema.statics.updateByEntry = function(entry) {
-  const { title } = entry
-  const at = entry.timestamp || Date.now()
-  const status = entry.timestamp === null ? 'down' : 'up'
+  const { title, timestamp: at = Date.now() } = entry
+  const status = parseInt(entry.timestamp, 10) > 0 ? 'up' : 'down'
 
   return this.findOneAndUpdate(
     { title: entry.title },
@@ -82,6 +79,7 @@ module.exports = class MongoStore {
 
   async init() {
     if (this.ready) return
+    // connect to database with crowi's configuration
     await crowi.setupDatabase()
     this.ready = true
   }
--- a/migrate-mongo-store.js
+++ b/migrate-mongo-store.js
@@ -1,0 +1,138 @@
+/**
+ * Crowi::migrateStore
+ *
+ * @package Crowi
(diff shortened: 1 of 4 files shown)


migrate-mongo-store.js
```diff
@@ -0,0 +1,138 @@
...
+      },
+      at: {
+        type: Date,
+        required: true,
+      },
+    },
+  ],
+})
+migrationSchema.methods.toEntry = function() {
+  const m = this
```
Sw0rdstream: unnecessary `this`

migrate-mongo-store.js
```diff
@@ -0,0 +1,138 @@
...
+  const entry = {
+    title: m.title,
+    timestamp: m.status === 'up' ? m.at.getTime() : null,
+  }
+  return entry
```
Sw0rdstream: literal variable `entry` is not accessed as left value. So the declaration is not needed.

just
```
return {
    title: this.title,
    timestamp: this.status === 'up' ? this.at.getTime() : null,
  }
```

package-lock.json
```diff
@@ -4373,7 +4373,7 @@
         },
         "string_decoder": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
+          "resolved": "http://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
```
Sw0rdstream: This line is changed

migrate-mongo-store.js
```diff
@@ -0,0 +1,138 @@
...
+}
+migrationSchema.statics.updateByEntry = function(entry) {
+  const { title } = entry
+  const at = entry.timestamp || Date.now()
```
Sw0rdstream: can also use destructing like this:
```
const { title, at = Date.now(), timestamp } = entry
```

migrate-mongo-store.js
```diff
@@ -0,0 +1,138 @@
...
+  const status = entry.timestamp === null ? 'down' : 'up'
```
Sw0rdstream: if timestamp is undefined or 0, status will be `up`. Is that your intention?

migrate-mongo-store.js
otofune:
```js
const { title, timestamp: at = Date.now() } = entry
```

migrate-mongo-store.js
otofune: I saw `tj/node-migrate`s implement: https://github.com/tj/node-migrate/blob/master/lib/migrate.js#L56-L57
I want to track 'down' event to know execution from another managers, but there are no hint without timestamp from store 😭

migrate-mongo-store.js
Sw0rdstream: my bad

migrate-mongo-store.js
otofune: I see, I'll use it! Thank you 🙏

PR #406, opened 2018-12-28
Rebuild index with zero downtime

lightnet328:
# Overview
By using aliases, we can rebuild indices with zero downtime.


lib/util/search.js
```diff
@@ -130,31 +134,94 @@ SearchClient.prototype.parseUri = function(uri) {
...
+  if (existsAlias) {
+    const aliases = this.client.cat.aliases({ name: this.indexNames.current, format: 'json' })
+    if (aliases.length > 0) {
+      return aliases[0]
+    }
+  }
+  return null
+}
+
+SearchClient.prototype.putAlias = async function(index) {
```
sotarok: 💬IMO: this `putAlias()` func name is good to be `putAliasOfCurrent()` and so far, because it is only used for adding `lcurrent` alias.

lib/util/search.js
```diff
@@ -130,31 +134,94 @@ SearchClient.prototype.parseUri = function(uri) {
...
+  return this.client.indices.putAlias({ index, name: this.indexNames.current })
+}
+
+SearchClient.prototype.ensureAlias = async function() {
+  const exists = await this.existsAlias()
+  if (!exists) {
+    const indices = await this.getIndices()
+    if (indices.length > 0) {
```
sotarok: 📝memo: If search index is never built yet, this gonna false and this function returns `false`.

PR #407, opened 2018-12-28
Add dnscache support

otofune:
To prevent slow dns resolution in vm on linux.
In December 2018, linux kernel may have race in conntrack.
See: https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts

At crowi, this problem causes very slow responses from `/files/:id`.


otofune:
I'm wondering ...
+ Is it better to use form to determine whether to enable dnscache?
  I used environment variables for it. This workaround is required by environment reason, so I think this choice is valid.
  But, I don't know theory...

PR #415, opened 2019-01-25
Suggest recently viewed pages

lightnet328:
# Overview
Suggest recently viewed pages in search suggests.

# Changes

- Behavior when the form is focused and the keyword is empty
    - Recently viewed pages are displayed up to 5 cases

- How to close search suggestions
    - Click outside the form and suggests

# Screens

localhost_3000_


sotarok:
❓ Is it deliberated that `userPageShow()` doesn't call `lru.add` so user page wouldn't appear on the recently viewed page list.

lib/service/lru.js
```diff
@@ -0,0 +1,55 @@
+class LRU {
+  constructor(crowi) {
+    this.crowi = crowi
+    this.max = 10
+    this.client = null
+
+    if (this.crowi.redisOpts) {
+      const redis = require('redis')
+      const redisClient = redis.createClient(this.crowi.redisOpts)
+      this.client = redisClient
```
sotarok: ❓ Is there any reason why you didn't use `crowi.redis` client?

lightnet328: No reason in particular. I change this service to reuse a connection.

lightnet328:
Thank you for your QA. I think those causes of bugs are follows.

1. The redirect link remains as Page.
2. LRU cache stores 10 IDs, but only 5 IDs were fetching. After that, IDs that do not exist are removed.

I'll fix `2` immediately. And I'll consider how to fix `1`.

sotarok:
@lightnet328 How is this question? https://github.com/crowi/crowi/pull/415#issuecomment-505488162

lightnet328:
> Is it deliberated that `userPageShow()` doesn't call `lru.add` so user page wouldn't appear on the recently viewed page list.

It's not deliberate. I change recently viewed page list to showing user pages.

PR #426, opened 2019-03-06
fix: correct shortpath for portals

sotarok:
currently, `shortpath()` for portal path doesn't work correctly.

### DIFF
--- a/client/components/PageList/PagePath.js
+++ b/client/components/PageList/PagePath.js
@@ -21,7 +21,7 @@ export default class PagePath extends React.Component {
     }
 
     // ページの末尾を拾う
-    const suffix = name.replace(/.+\/(.+)?$/, '$1')
+    const suffix = name.replace(/.+\/(.+)$/, '$1')
     return suffix || name
   }
 


lib/util/middlewares.js
```diff
@@ -74,7 +74,7 @@ exports.swigFilters = function(app, swig) {
       }
 
       // ページの末尾を拾う
-      const suffix = name.replace(/.+\/(.+)?$/, '$1')
+      const suffix = name.replace(/.+\/(.+)$/, '$1')
```
lightnet328: `path2name` filter is equivalent to `getShortPath` method. So this method should be changed.

In the future, I believe that we can delete `path2name` filter by replacing Swig templates with React.
For that reason, I've copied and pasted as a temporary measure.

I can use Suggest for changes, but `middlewares.js` was not changed in this pull request, so I committed directly.
https://help.github.com/en/articles/reviewing-proposed-changes-in-a-pull-request

sotarok: Oh, thanks!

Issue #344, opened 2018-09-10
Search from pages which created by myself


sotarok:

sotarok:
Implemented #427 and will be released on 1.8.
And also duplicated #260

PR #427, opened 2019-03-22
Search from pages which created by myself

lightnet328:
# Overview
1. Add pages that are not public in Elasticsearch
2. Change search results by `grant` property and `username` property

#344


lightnet328:
Yes, I've resolved conflicts.

lib/util/search.js
```diff
@@ -307,6 +303,7 @@ SearchClient.prototype.prepareBodyForCreate = function(body, page, index = null)
     path: page.path,
     body: page.revision.body,
     username: page.creator.username,
+    grant: page.grant,
```
sotarok: @lightnet328 It seems the schema of doc on es has been changed.
So should we apply this change to `mappings.json` ?

lightnet328: I think so too. I'll add grant field to `mapping.json`.

PR #433, opened 2019-05-10
chore: Add Redirect URLs help to how to setup Slack Integration

potato4d:
## Overview

I added the explanation about Redirect URL to the explanation flow of Slack integration.

It is necessary setting to enable Slack integration, But it's description is not exist in `How to configure Slack app for Crowi`.

Thanks.

### Before

Screen Shot 2019-05-10 at 15 30 51

### After

Screen Shot 2019-05-10 at 15 29 02


sotarok:
Thank you so much. It is so helpful :)

PR #443, opened 2019-05-20
move ci config to designated location

inductor:
Since CircleCI 2.0, it is recommended to put the config file in `.circleci/config.yml` so this is the PR for it.


inductor:
Huh, the test failed but I do not think it's because of the PR because it doesn't change anything its codebase 🤔

inductor:
Succeeded! Yay 🎉

lightnet328:
It's just timeout. I rebuild this changes.

Issue #445, opened 2019-05-29
Optimize docker image


inductor:
At the moment the crowi docker image size is over 1GB(more precisely 1.16GB).

I think it can be shrunk by doing the followings.

- Asset compile should be done before the server daemon starts.
	- Use multi-stage build so that you won't need `devDependencies` in production build.
- Use alpine node image for production running.
	- For asset compiling you can still use the debian based image.
- untrack unnecessary files in .dockerignore
	- .git
	- pre-complied asset files
	- test codes
	- other setting files such as elastic search and .circleci

inductor:
I can contribute with the docker-wise problems, but I will probably need your help for asset compiling.

lightnet328:
Thank you for the great advice! I created a PR related to this issue.
I didn't ignore test codes and setting files. This is because I think text files like codes and settings size are not so big.
Please review this PR, if it's OK with you.

inductor:
Oops I closed the issue by mistake

Issue #410, opened 2019-01-21
Support Node.js v10?


inductor:
It's been a while since Node.js 10.x became stable. Any plan to support v10?

Thanks!

sotarok:
Crowi v1.8 will use v10.x LTS :)

PR #446, opened 2019-05-30
Optimize docker image

lightnet328:
# Overview

See #445.
Thank you for creating issue easy to understand, @inductor.

# Changes

| | Commit | Size |
|:--|:--|:--|
| Before | | 1.43GB |
| Use Multi-stage build | c093783 | 729MB |
| Ignore .git | 8445930| 540MB |

These images are built for development.
In production, `mongodb-memory-server` is not installed. So the production image will be more lightweight.


lightnet328:
memo: I want to upgrade Node.js version to 10.16 after this PR merged.
https://github.com/crowi/crowi/issues/410

inductor:
Related issue: https://github.com/crowi/crowi/issues/445

.dockerignore
```diff
@@ -1,2 +1,3 @@
 data
-node_modules
\ No newline at end of file
+node_modules
+.git
```
inductor: Please put new line at EOF for POSIX compatibility.

http://hiroakiuno.hatenablog.com/entry/20070409/p1
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

inductor:
The image size became 530MB from 1.16GB ❤️

.dockerignore
inductor: @lightnet328 ping

lightnet328: Thanks. I agree with that and fixed it.

PR #449, opened 2019-06-05
Change from alpine image to stretch-slim

lightnet328:
# Background
We changed the base image of Docker from Debian Stretch to Alpine. #446
As MongoDB does not provide binary files for Alpine, mongodb-memory-server has stopped working.

# Changes
To fix this problem, change the base image to a lightweight Debian Stretch image (stretch-slim).
This change will increase the size of the image by about 100 MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.12.0-alpine as builder
+FROM node:10.16.0-stretch-slim as builder
 
 ARG NODE_ENV="production"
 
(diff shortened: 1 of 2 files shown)


sotarok:
At this moment, I approved this change but I'd like to think how to use alpine image again if we'll come up with an alternative idea to run mongodb-memory-server.

inductor:
https://github.com/franz-josef-kaiser/Docker-Alpine-Nodejs/issues/13 D:

PR #435, opened 2019-05-18
Ensure config consistency on multi apps

lightnet328:
# Overview
Servers have config cache and those config are not synced.
Therefore, consistency in config can not be ensured on multiple apps environment.

# Changes

Sync config using Redis PubSub when config changed on other server.

# Test

1. Edit `docker-compose.development.yml`

```yml
...
  crowi:
...
    ports:
      - 3000-3001:3000
...
```

2. `docker-compose -f docker-compose.development.yml --scale crowi=2`

3. Browse `localhost:3000` and change config

4. Browse `localhost:3001`

## Related Issues

- Close #8

PR #451, opened 2019-06-06
Reload fields depends on config even when config updated in another node

otofune:
Overview
---
#8 was resolved by https://github.com/crowi/crowi/pull/435. But it doesn't include reloading some instances that depends on config.
This PR adds it.

### DIFF
--- a/lib/crowi/index.js
+++ b/lib/crowi/index.js
@@ -136,11 +136,12 @@ Crowi.prototype.setupConfigPubSub = function() {
     debug('PubSubId', this.pubSubId)
 
     if (this.subscriber) {
-      this.subscriber.on('message', (channel, message) => {
+      this.subscriber.on('message', async (channel, message) => {
         if (channel !== this.configPubSubChannel) return
         const { pubSubId, config } = JSON.parse(message)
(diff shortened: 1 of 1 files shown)


lib/crowi/index.js
```diff
@@ -136,11 +136,12 @@ Crowi.prototype.setupConfigPubSub = function() {
     debug('PubSubId', this.pubSubId)
 
     if (this.subscriber) {
-      this.subscriber.on('message', (channel, message) => {
+      this.subscriber.on('message', async (channel, message) => {
         if (channel !== this.configPubSubChannel) return
         const { pubSubId, config } = JSON.parse(message)
         if (pubSubId === this.pubSubId) return
         this.config = config
+        await Promise.all([this.setupSlack(), this.setupMailer()])
```
otofune: [note] I thought that it's bit difficult to detect what instances to reload by difference from current config. Also it will confuse someone reading this.
This is why of adding this only.

otofune: Also I checked that there are no problem to call `setupSlack` and `setupMailer` always.

PR #481, opened 2019-07-07
Fix recently viewed pages api endpoint

lightnet328:


lightnet328:
We will make more general discussions about endpoint naming conventions.

PR #486, opened 2019-07-22
Convert backend layers to TypeScript

otofune:
- Import https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36884


lib/crowi/index.ts
```diff
@@ -134,7 +168,7 @@ class Crowi {
 
   // getter/setter of model instance
   //
-  model(name, model) {
+  model(name: any, model?: any) {
```
hiroppy: I think we can change to `string` and `number`.(or only string)

otofune: fixed @ https://github.com/crowi/crowi/pull/486/commits/da2644158198183c41be09c25e6d25efdef7dd49

lightnet328:
I took over this PR from @otofune 🙋‍♂️

PR #487, opened 2019-07-24
Add the first case of express test.

sotarok:
Changes and reasons:

- Initialize Crowi object for each test.
    - To run test routes and controllers.
    - And I added versions.get API and its test for the first case.
- Crowi object
    - Changed because `start()` method was connected with setup of express. Separate settings to `init()` and run (listen) to `start()`
    - And because of this change, some method, like events, had to changed because it assumed that the `crowi.io` is always existing but `io` is not set if called only `init()` without called `start()` from now on.
- BASE_URL
    - Backlink models sometimes brokes.
- `--maxWorkers` set to test
    - Because a mysterious error `Topology was destroyed` happened (It might be because of CircleCi's environment).
        - Finally, I couldn't find the actual reason why this error happened but I found the workaround to avoid this error, that is run with maxWorkers (or runInBand)
    - https://circleci.com/gh/crowi/crowi/1917 (while trying)
- Remove workaround codes in tests that is added because Crowi object was not initialized.
- Fix lint for client files.
    - Because I updated eslint, many of client files had to be fixed.


.circleci/config.yml
```diff
@@ -5,27 +5,25 @@ jobs:
...
-          NODE_ENV: development
-          MONGO_URI: mongodb://localhost:27017/crowi
           REDIS_URL: redis://localhost:6379/crowi
-          ELASTICSEARCH_URI: //localhost:9200/
           PLANTUML_URI: //localhost:8080
           FILE_UPLOAD: local
           MONGOMS_DOWNLOAD_MIRROR: https://downloads.mongodb.org
-      - image: mongo:3.6.13
+          MONGOMS_DISABLE_POSTINSTALL: 1
+      - image: circleci/redis:5
```
lightnet328:
```suggestion
      - image: circleci/redis:5.0.5
```

I would like to specify a minor version to change the version explicitly.

test/models/backlink.test.js
```diff
@@ -1,42 +1,58 @@
...
-  const conn = utils.mongoose.connection
-  const appUrl = 'http://localhost:3000'
+  let Backlink
+  let Page
+  let Revision
+  let conn
   let user
 
-  beforeAll(async () => {
+  beforeAll(done => {
```
lightnet328: I think calling `done()` is unnecessary for sync function.

test/models/backlink.test.js
```diff
@@ -1,42 +1,58 @@
...
+    Backlink = crowi.model('Backlink')
+    Page = crowi.model('Page')
+    Revision = crowi.model('Revision')
+    conn = crowi.getMongo().connection
+
+    done()
+  })
+
+  beforeAll(async done => {
```
lightnet328: It's also unnecessary. I think `done()` should be used only for processing in asynchronously using a callback.

test/util/slack.test.js
```diff
@@ -1,21 +1,10 @@
...
 describe('Slack Util', () => {
-  const crowi = new (require(ROOT_DIR + '/lib/crowi'))(ROOT_DIR, process.env)
-  const slack = require(crowi.libDir + '/util/slack')(crowi)
+  let slack
+  beforeAll(() => {
+    slack = require(crowi.libDir + '/util/slack')(crowi)
+  })
 
   test('convert markdown', () => {
-    // FIXME: Mock. Should be changed to use a real Crowi object.
```
lightnet328: Nice👍

PR #490, opened 2019-08-05
Compile backend TypeScript sources

otofune:
Heroku free dyno can't boot with `ts-node` because compiling timed out.
This PR resolves it.


otofune:
TODO
---
- [ ] Verified docker image works well in my local machine.
- [ ] Verified Heroku App works well.
- [ ] Confirm that there are no problem to move directory and files

package.json
```diff
@@ -175,30 +174,34 @@
...
     "dev": "concurrently --kill-others 'npm run dev:server' 'npm run dev:client' 'npm run watch'",
     "dev:server": "DEBUG=crowi:* NODE_ENV=development nodemon --watch ./lib/ --watch ./locales/ --ext 'ts,tsx,js,jsx,json' --exec npm run start",
     "dev:client": "NODE_ENV=development webpack --progress --profile --colors --display-error-details --watch --config tools/webpack.config.js",
     "format": "eslint --fix '**/*.{ts,tsx,js,jsx}'",
     "lint": "eslint '**/*.{ts,tsx,js,jsx}'",
     "test": "DEBUG= jest --maxWorkers=5",
     "coverage": "npm test -- --coverage",
-    "build": "npm run prod:client",
+    "build": "run-s build:*",
+    "build:server": "babel --no-babelrc --config-file ./.server.babelrc lib --out-dir dist --extensions '.ts'",
```
otofune: Why using babel, I want to use `module-resolver` because `tsc` doesn't resolve `paths` to relative path.

lib/app.ts
```diff
@@ -7,11 +7,12 @@
 
 import dotenv from 'dotenv'
 import Crowi from 'server/crowi'
```
otofune: Why to move this file to `lib/`, `babel` can't build file & directory one time and `module-resolver` can't resolve 'server/' as `./dist`.

---

もう一つ言うとすれば、置くなら dist/ 内に入れたいというのがありました。その時はディレクトリ構造から rootDir を L15 のようにする必要がありますが、app.ts をそのまま `ts-node` で動く状態を維持しつつ path.join(__dirname, `..`) することはできないのでこうしました。

Dockerfile
```diff
@@ -1,21 +1,23 @@
...
 
 ADD . .
+RUN npm run build
 
-# Run postinstall manually
-RUN npm run postinstall
+RUN rm -rf lib client
 
+# Remove devDependencies if NODE_ENV is production
+# TODO: verify that crowi can boot normally without devDependencies
```
otofune: I'll try this after this PR.

.server.babelrc
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./.babelrc",
+  "presets": [
+    ["@babel/env", {
+      "targets": {
+        "node": "current"
```
otofune: I didn't want `regenerate-runtime`, so I splitted babelrc. 
But this is required to change `import` to `require`.

lib/crowi/index.ts
```diff
@@ -53,8 +54,6 @@ class Crowi {
 
   publicDir: string
 
-  libDir: string
```
otofune: At this moment, no one requires this variable.

```sh
$ git grep libDir
$
```

package.json
```diff
@@ -175,30 +174,34 @@
...
+    "build:client": "npm run prod:client",
     "prod:client": "NODE_ENV=production webpack --bail --colors --display-error-details --config tools/webpack.config.js",
     "type-check": "tsc --noEmit",
     "type-check:watch": "npm run type-check -- --watch",
     "webpack": "webpack",
-    "watch": "onchange '**/*.{ts,tsx,js,jsx}' -e 'bin' -e 'client/thirdparty-js' -e 'node_modules' -e 'public' -f add -f change -- eslint --fix '{{changed}}'",
-    "postinstall": "sh -c '[ \"$SKIP_POSTINSTALL\" != \"\"  ] || npm run build'"
```
otofune: postinstall was required for Heroku maybe. But Heroku can use `build` script at this moment.

PR #497, opened 2019-08-07
Modify dockerfile

inductor:
- Use `COPY` rather than `ADD` for better caching
  - ADD works better when fetching a file from URL
- Use [] for CMD as it's [recommended](https://docs.docker.com/engine/reference/builder/#cmd)

I'm also planning to make a PR for updating node to 10.16.2 but it's still [on the way](https://github.com/nodejs/docker-node/pull/1085)


sotarok:
Thank you for your contribution ;)

PR #510, opened 2019-08-08
Introduce tj/migrate

otofune:
Recrate #401 because closed accidentally by target branch was removed.


otofune:
If this is required again, I will re-create PR

PR #523, opened 2019-08-14
Feature new header

### LINKED ISSUE #488: Appear [x] of search form at unexpected position on Firefox
It is really small bug but anyway reported.

Windows10
Firefox 68.0.1

image

<details>
    <summary>Japanese</summary>
とてもとても軽微な問題ですが、Issueとして登録させていただきます。
</details>

sotarok:
Implemented header style change:

# New white styled header

image

notification:
image

user menu:
image

# New breakpoint

`md` and `sm` have new mobile menu bars:

image

menu opened:
image

# New page list DOM structure

image

Originally, it is just a list for icon and page path so if pages have long name, it was shown as a line with icon **AND** path like below:
image

Now icon and page path in page list has a new structure to fix it :)

----

also close #488


suzuki:
I don't know whether this change is the cause or not, but it looks like this:

スクリーンショット 2019-08-16 17 37 28

The showing the empty box is not expected maybe.

resource/css/_page_list.scss
```diff
@@ -14,6 +14,9 @@
   }
 
   .page-list-ul {
+    width: calc(100vw - 2em);
```
lightnet328: Because of `.page-list-ul` is also used in the search result, it should be removed.

sotarok: fixed the style and also adjust adjust small device style.

PR #470, opened 2019-06-30
Refactor routes and middlewares

lightnet328:
# Overview

Routers and middleware are becoming more complex.
I changed routes and middlewares to achieve the following two goals:

- Facilitate parallel development
- Improve file readability

# Changes

- Split middlewares per modules
- Move middleware module files from the util directory to the middleware directory
- Split routes per domains
- Move controllers files to the controller directory

# Before

```
util
└── middleware.js
```

```
routes *
└── index.js
```

* Other files are essentially controllers.

# After

```
middlewares
├── accessTokenParser.js
├── adminRequired.js
├── applicationInstalled.js
├── applicationNotInstalled.js
├── awsEnabled.js
├── basicAuth.js
├── clientContext.js
├── csrfVerify.js
├── encodeSpace.js
├── fileAccessRightOrLoginRequired.js
├── i18next.js
├── index.js
├── loginChecker.js
├── loginRequired.js
├── swigFilters.js
└── swigFunctions.js
```

```
routes
├── api
│   ├── admin.js
│   ├── attachment.js
│   ├── bookmark.js
│   ├── comment.js
│   ├── index.js
│   ├── like.js
│   ├── notification.js
│   ├── page.js
│   ├── revision.js
│   └── share.js
├── admin.js
├── index.js
├── login.js
└── me.js
```

```
controllers
├── admin.js
├── attachment.js
├── backlink.js
├── bookmark.js
├── comment.js
├── index.js
├── installer.js
├── login.js
├── logout.js
├── me.js
├── notification.js
├── page.js
├── revision.js
├── search.js
├── share.js
├── shareAccess.js
├── slack.js
└── user.js
```

PR #531, opened 2019-08-18
Separate login and register pages and update its style

sotarok:
Apply new styles (base branch is #523 )

## New login page style

- New background image.
- Flex based vertical centralize.

image

## Separate login/register page and remove flipper 

- /login and /register are independent.

## Fix bug of redirect loop on /me/auth/third-party

- In loginRequired middleware, it determined if it is auth page or not with `req.path` but caused by refactoring of router structure (#470) the `path` is now returning only the path matches to the router definition itself.
- So use originalUrl to determine it.


views/register.html
```diff
@@ -0,0 +1,124 @@
+{% extends 'layout/single-nologin.html' %}
+
+{% block html_title %}{{ t('page_login.sign_in') }} 揃 {% endblock %}
```
lightnet328: 揃 👈

sotarok: 😭 
fixed.

PR #537, opened 2019-08-18
Update i18next

sotarok:
- Bump versions of i18next related libs together.
- Close #493, #524, #533, #555


otofune:
:eyes: https://github.com/i18next/i18next/blob/master/CHANGELOG.md#1709

PR #539, opened 2019-08-18
Fix search suggest styles

lightnet328:
# Changes

- Stretch the search suggest to maximum on desktop
- Show loading and no result message on the recently viewed pages list

# Screens

## Loading

localhost_3000_ (2)

## Loaded

localhost_3000_ (1)

## No results

localhost_3000_

# Related

https://github.com/crowi/crowi/pull/523#issuecomment-521934843

### DIFF
--- a/resource/css/_search.scss
+++ b/resource/css/_search.scss
@@ -34,7 +34,7 @@
     width: 100%;
 
     // for SearchSuggest
-    @include media-breakpoint-up(md) {
+    @include media-breakpoint-up(lg) {
       position: relative;
     }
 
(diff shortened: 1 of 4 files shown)


resource/css/_search.scss
```diff
@@ -87,15 +92,13 @@
 }
 
 .search-suggest {
-  max-width: 500px;
+  width: 100%;
```
sotarok: If here to fix value to `100%`, small width device displays really narrow area.
So, should we specify `min-width` or the style between `md` and `lg`?

For example, below is around 800px width window.

image

PR #549, opened 2019-08-22
Introduce Storybook

saitoeku3:
## Overview

We can see components in developing on browser using Storybook.
So I thought Storybook would help newcomers with #452 and added it.
If you agree with my opinion, please consider to introduce Storybook.

## Changes

- Add Storybook settings corresponding to tsx
- Use `:6001` for Storybook
- Start with `docker-compose -f docker-compose.development.yml exec crowi npm run storybook`

## Screenshot

This is `client/components/Common/Modal` on Storybook.
I feel it doesn't has its own styles. Is it displayed correctly?

Screen Shot 2019-08-23 at 0 54 16

### DIFF
--- a/client/components/Common/Icon.stories.tsx
+++ b/client/components/Common/Icon.stories.tsx
@@ -1,0 +1,8 @@
+import React from 'react'
+import styled from 'styled-components'
+import { storiesOf } from '@storybook/react'
+import Icon from './Icon'
+
+storiesOf('Common/Icon', module)
+  .add('Default', () => <Icon name="help-circle" spin={false} />)
+  .add('Spin', () => <Icon name="help-circle" spin={true} />)
--- a/client/components/Common/Icon.stories.tsx
+++ b/client/components/Common/Icon.stories.tsx
@@ -1,5 +1,4 @@
 import React from 'react'
-import styled from 'styled-components'
 import { storiesOf } from '@storybook/react'
 import Icon from './Icon'
 
--- a/client/components/Common/Icon.stories.tsx
+++ b/client/components/Common/Icon.stories.tsx
@@ -3,5 +3,5 @@ import { storiesOf } from '@storybook/react'
 import Icon from './Icon'
 
 storiesOf('Common/Icon', module)
-  .add('Default', () => <Icon name="help-circle" spin={false} />)
+  .add('Default', () => <Icon name="help-circle" />)
   .add('Spin', () => <Icon name="help-circle" spin={true} />)
--- a/client/components/Common/Modal.tsx
+++ b/client/components/Common/Modal.tsx
@@ -9,7 +9,7 @@ export default class Modal extends React.Component<{}, State> {
     super(props)
 
     this.state = {
-      modalShown: true,
+      modalShown: false,
     }
   }
 
--- a/client/components/Common/Modal.tsx
+++ b/client/components/Common/Modal.tsx
@@ -9,7 +9,7 @@ export default class Modal extends React.Component<{}, State> {
     super(props)
 
     this.state = {
-      modalShown: false,
+      modalShown: true,
     }
   }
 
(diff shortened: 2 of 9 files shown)


sotarok:
@saitoeku3 Thank you for your great contribution.
That's awesome!
Can you rebase latest `master` first?

client/components/Common/Modal.tsx
```diff
@@ -9,7 +9,7 @@ export default class Modal extends React.Component<{}, State> {
     super(props)
 
     this.state = {
-      modalShown: false,
+      modalShown: true,
```
saitoeku3: I changed `modalShown` to preview on Storybook.
If there are some problems, please tell me!

lightnet328: Uh, it's no longer used. Would you remove this component and use `UserDate` instead of it?

stories/Common/Modal.stories.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
```
lightnet328: Would you move this component to `client/components/Common/Modal.stories.ts`?
I'd like to place story files at the same level as components.

package.json
```diff
@@ -199,6 +201,7 @@
     "dev:client": "NODE_ENV=development webpack --progress --profile --watch --config webpack/client/webpack.config.js",
     "format": "eslint --fix '**/*.{ts,tsx,js,jsx}'",
     "lint": "eslint '**/*.{ts,tsx,js,jsx}'",
+    "storybook": "start-storybook --ci -p 6001",
```
lightnet328: I'd like to run `storybook` script when run `dev` script. What do you think?

saitoeku3: I overlooked `dev` script… It's a great idea!!!

stories/Common/Modal.stories.tsx
saitoeku3: I sympathize your idea.
So I move a story file to the same level as components.

client/components/Common/Modal.tsx
saitoeku3: Does it mean I would completely remove this component from this app or only a story file?

saitoeku3: @lightnet328 Should I remove `client/components/Common/Modal.tsx` ?

sotarok: @saitoeku3 It seems `Common/Modal` component added as trial of replacement from jQuery to React in past (because the value of form is fixed and it clearly not used), and actually, it is not `no longer used` but `not yet used` I think.
I'm sorry that it confuses you because of the incompleted component.

How about try `Icon` as an initial example to introduce Storybook?
`Icon` is the good simple component.

saitoeku3: @sotarok  I did it! Thanks for your good suggestion!

Screen Shot 2019-08-27 at 23 41 06

Aug-27-2019 23-42-59

client/components/Common/Icon.stories.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import styled from 'styled-components'
+import { storiesOf } from '@storybook/react'
+import Icon from './Icon'
+
+storiesOf('Common/Icon', module)
+  .add('Default', () => <Icon name="help-circle" spin={false} />)
```
lightnet328: nits:
As a default example, I think it should be described **only** required props.
```suggestion
  .add('Default', () => <Icon name="help-circle" />)
```

client/components/Common/Icon.stories.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import styled from 'styled-components'
```
lightnet328: nits:
Please remove unused import.
Sorry, I have should add a `import/no-unused-modules` rule to ESLint.

```suggestion
```

saitoeku3: Sorry, I forgot removing this line…

PR #557, opened 2019-08-30
Fix: admin page bug

sotarok:
- Fix: Can't update aws setting
- Fix: Unexpected quoted HTML in locale
- Fix: Search index name in admin page


client/components/Admin/App/AWSSettings.tsx
```diff
@@ -16,18 +16,18 @@ interface Props {
 
 const AWSSettings: FC<Props> = ({ settingForm, update, alert }) => {
   const [t] = useTranslation()
-  const [region, setRegion] = useState(settingForm['aws:region'])
-  const [bucket, setBucket] = useState(settingForm['aws:bucket'])
-  const [awsAccessKeyId, setAwsAccessKeyId] = useState(settingForm['aws:awsAccessKeyId'])
-  const [awsSecretAccessKey, setAwsSecretAccessKey] = useState(settingForm['aws:awsSecretAccessKey'])
+  const [region, setRegion] = useState(settingForm['aws:region'] || '')
+  const [bucket, setBucket] = useState(settingForm['aws:bucket'] || '')
+  const [accessKeyId, setAccessKeyId] = useState(settingForm['aws:accessKeyId'] || '')
```
sotarok: Server returns not `aws:awsAccessKeyId` but `aws:accessKeyId`.

client/components/Admin/TopPage.tsx
```diff
@@ -47,8 +47,7 @@ const TopPage: FC<{}> = () => {
           {/* TODO: multiple nodes */}
           {searchConfigured ? (
             <>
-              Configured: {host}
-              {indexName}, <strong>{esVersion}</strong>
+              Configured: {host}/{indexName}, <strong>{esVersion}</strong>
```
lightnet328: Maybe `host` or `indexName` is including `/` already.

sotarok: No, actually not.

It parsed and separated to host and indexName
https://github.com/crowi/crowi/blob/e455d11ec5b29a17a7ee8b6d8fabe54a4f023c30/lib/util/search.ts#L125

sotarok: Hm.. in case of no index name specified?

lightnet328: Oh, that's it!
In my environment, `ELASTICSEARCH_URL` is `elasticsearch:9200/`.

sotarok: Okay... I see.
Anyway, both host and indexName spec should be fixed even any env passed.

sotarok: I confirmed `{host}` not includes `indexName`.

image

I added the test if the index name not specified and set `crowi` as default.
https://github.com/crowi/crowi/pull/557/commits/b97b12c21ed2ca3f688565dd131d0525b0a759d2#diff-a51ffab4bb0b2ada968f2932a43d1497R23

lightnet328: Ah, I think the cause is different from this.
The true cause is that passed `uri` is not matched with `/^(https?:\/\/[^/]+)\/(.+)$/`.
So matched string is assigned to `host` and `indexName` and these remains at the initial values.

sotarok: Because Crowi uses the ES URL to determine index name with parsing it.
So the protocol should be specified.
(In addition, the es client adds the defaultProtocol if it is no specified https://github.com/elastic/elasticsearch-js-legacy/blob/16.x/src/lib/host.js#L86 )

So I changed below:
- Error when the protocol not specified.
- And use [WHATWG API](https://nodejs.org/docs/latest/api/url.html#url_whatwg_api) to parse it.

lib/util/search.ts
```diff
@@ -118,12 +118,15 @@ SearchClient.prototype.shouldIndexed = function(page) {
 // BONSAI_URL is following format:
 // => https://{ID}:{PASSWORD}@{HOST}
 SearchClient.prototype.parseUri = function(uri) {
+  if (!uri.startsWith('http')) {
+    throw new Error('URL for Elasticsearch shuold starts with http/https')
```
lightnet328: [nits]
`shuold` should be changed to `should`.
```suggestion
    throw new Error('URL for Elasticsearch should starts with http/https')
```

client/components/Admin/TopPage.tsx
lightnet328: Thanks for your description. Okay, it's LGTM.

PR #563, opened 2019-09-03
Fix issue admin notification form values and newly fetched values are not synced

lightnet328:
# Overview

Currently, forms keep using initial values. I fixed this.


lightnet328:
> React shows warnings below on notification settings screen. Can you reproduce it?
> 
> ```
> Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
> ```

Yes, I just fixed this problem.

PR #579, opened 2019-09-07
Dockerfile: Add procps (ps)

otofune:
nodemon (used in `dev:server` script) can't boot without ps command.

I didn't notice, but sometime `ps` command left from Docker image.
Please re-build image before testing this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,10 @@ ARG NODE_ENV="production"
 ENV CROWI_VERSION v1.8.0
 ENV NODE_ENV ${NODE_ENV}
 
+# 'ps' command is missing in node:10.16.2-stretch-slim
+# but required by pstree.remy (used from nodemon)
+RUN apt-get update && apt-get install -y procps && apt-get clean
+
 USER node
 
 WORKDIR /crowi


otofune:
:memo: 
https://github.com/remy/nodemon/blob/5124ae9528da8b19ce25c1945d1e62c35f20e861/lib/monitor/run.js#L15
https://github.com/remy/pstree/blob/d3d24031e3c7547e46f0aba3dd4e439a5e5d1665/lib/index.js#L7-L9

lightnet328:
What makes you think so?

I'm not familiar with this, but I think `pstree` really doesn't require `ps` command.
It seems to get process ids using other commands when `ps` command doesn't exist.
https://github.com/remy/pstree/blob/d3d24031e3c7547e46f0aba3dd4e439a5e5d1665/lib/index.js#L20

otofune:
Sorry, it's my fault... In my computer log showed `ps` command execution error, so I believed that pstree associated with `ps` strongly.

I re-built docker image in master branch, and it works at now.
So I closes this... :bow:

lightnet328:
I see. Thank you for the explanation👍

PR #581, opened 2019-09-08
Move view files to lib dir

lightnet328:
# Overview
Previously, views files were moved to the project's root dir for fixing a problem related to compiling TypeScript in https://github.com/crowi/crowi/pull/490.
And now, we can move view files to lib dir because of compiling method of TypeScript is changed.

Views files are only handled by express. So I think view files had better be placed on lib dir in my opinion. Please tell me your opinions if you have any 🙂


lightnet328:
Ah..., I think most likely this branch conflict with migrating to SPA. So I close this PR.

PR #595, opened 2019-09-13
Divide AWS settings by domain

lightnet328:
# Overview

Divide AWS settings according to usage. SES doesn't support many regions compared to S3, so it's more convenient if an administrator can configure them separately.

# Screens

localhost_3000_admin_app(iPad Pro)


locales/en-US/translation.yml
```diff
@@ -259,15 +258,22 @@ admin:
     client_secret: Client Secret
   mail:
     legend: Mail Settings
-    tips1: SMTPの設定がされている場合、それが利用されます。SMTP設定がなく、AWSの設定がある場合、SESでの送信を試みます。
+    tips1: SMTPの設定がされている場合、それが利用されます。SMTP設定がなく、AWSの設定がある場合、SESでの送信を試みます。SESを利用するためにはFromメールアドレスのVerify、プロダクション利用設定をする必要があります。
```
sotarok: 👍

lib/form/admin/mail.ts
```diff
@@ -9,4 +9,11 @@ export default form(
     .toInt(),
   field('settingForm[mail:smtpUser]', 'SMTPユーザー').trim(),
   field('settingForm[mail:smtpPassword]', 'SMTPパスワード').trim(),
+  field('settingForm[mail:aws:region]', 'リージョン')
+    .trim()
+    .is(/^[a-z]+-[a-z]+-\d+$/, 'リージョンには、AWSリージョン名を入力してください。 例: ap-northeast-1'),
```
sotarok: 🤔 
I know it is not an actual problem, but AWS has a region that not matched with this pattern like `us-gov-west-1` (Of course, `us gov` might not use Crowi though 😅)

https://docs.aws.amazon.com/ja_jp/general/latest/gr/rande.html

lib/models/config.ts
```diff
@@ -254,6 +323,26 @@ export default (crowi: Crowi) => {
     return localConfig
   }
 
+  configSchema.statics.migrate = async function() {
```
sotarok: Shouldn't the migration function check the current status which version is running previously like if `aws:bucket` exists means the version < v1.7 ?
It is a bit danger if always migration function run without any check IMO.

lib/form/admin/mail.ts
lightnet328: Ah, I see.

lib/models/config.ts
lightnet328: `Config.migrate` method is idempotent because of `Config.copyByParams` method checks if `oldKey` exists. So I think that checking `aws:bucket` is redundant.

By the way, in essence, it should manage the migration with a database.

locales/en-US/translation.yml
lightnet328: What's your point?

sotarok: I just wanted to say IINE!

lightnet328: Ah, I see😂

PR #602, opened 2019-09-27
Render all material design icons using SVG

lightnet328:
# What

- render all material design icons using SVG
  - use svg path directly instead of icon fonts
  - use icon component
- remove @mdi/font

# Why

- unify rendering icons method
- reduce the loading time of icon font


sotarok:
@lightnet328 Can you check the test failure?

PR #603, opened 2019-09-29
Add components stories

lightnet328:
# WHAT

- Add component stories
  - Existing components aren't changed

# WHY

- To be able to see components independently
- To make it easier to increase the independence of components in the future


sotarok:
@lightnet328 One thing ... how about ignore `*.stories.ts` from the coverage?
(increased red signals 😿 )

PR #608, opened 2019-10-05
Fix sticky page header and some styles

### LINKED ISSUE #562: Upside gap from header found in restricted page
scroll and then:

Screenshot from 2019-09-03 22-54-23

lightnet328:
# Changes

- Fix sticky page header

- Fix page grant styles

- Fix global menu styles and animation

- Fix page height to fit the screen height

Close #562


resource/css/_page.scss
```diff
@@ -146,18 +150,15 @@
   .main.grant-specified,
   .main.grant-owner {
     background: #333;
-    padding: 0 16px 16px 16px;
```
sotarok: Why removed this padding?

Original:
image


Current branch:
image


The style looks a bit broken ...

client/crowi.ts
```diff
@@ -604,9 +604,9 @@ $(function() {
     }
 
     // header
-    var $headerWrap = $('#page-header').parent()
+    var $headerWrap = $('.header-wrap')
     if ($headerWrap.length > 0) {
-      const headerHeight = $('.crowi-header').outerHeight() || 0
+      const headerHeight = $('.v2-crowi-header').outerHeight() || 0
```
sotarok: 👍 Nice fix.
But I found the sticky header broken in case the page is not public. Can you also fix it?

image

resource/css/_page.scss
lightnet328: In particular, I fixed this because the display width on smartphones become narrower.

lightnet328: I've just fixed broken styles. The modified styles looks like this:
localhost_3000_user_lightnet328_memo_2019_10_12 (1)

client/crowi.ts
lightnet328: Oh, I just fixed it.

PR #634, opened 2019-12-04
Fix page create modal and global menu

lightnet328:
# WHAT

- Fix an issue that page creation modal can not be opened from the navigation drawer
  - Accordingly, the navigation drawer was migrated to React component
  - Split PageCreateModal into button and modal
- Upgrade some packages for using optional chaining


client/components/NavigationDrawer/NavigationDrawerOpener.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import Crowi from 'client/util/Crowi'
+import Icon from 'components/Common/Icon'
+import NavigationDrawer from 'client/components/NavigationDrawer/NavigationDrawer'
+
+interface Props {
+  crowi: Crowi
+}
+
+const NavigationDrawerContainer: FC<Props> = ({ crowi }) => {
```
saitoeku3: This component's name is different from filename.
I think it should be same name `NavigationDrawerContainer` or `NavigationDrawerOpener`.

lightnet328: Ah, good catch! I renamed it to `NavigationDrawerContainer`.

PR #676, opened 2020-04-29
Update packages and Node.js version

lightnet328:
# WHAT

- Upgrade Node.js version to v12.16.2
- Update packages
  - Fix type errors
  - Format codes

Please see the essential diff from https://github.com/crowi/crowi/pull/676/files?file-filters%5B%5D=.json&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile.


lightnet328:
I think resolving these conflicts is more difficult than creating a new one. Please create a new PR when we work on upgrading to Node.js v12.

PR #681, opened 2020-05-15
Fix slack config and implement "remove slack config"

sotarok:
## What

- Load slack instance after slack config set.
- Implement "Remove slack config" feature.

Fix errors below:
image


lib/routes/api/admin.ts
```diff
@@ -26,6 +26,7 @@ export default (crowi: Crowi, app: Express, form) => {
 
   router.get('/admin/notification', Admin.api.notification.index)
   router.post('/admin/notification/slackSetting', csrf, form.admin.slackSetting, Admin.api.notification.slackSetting)
+  router.post('/admin/notification/removeSlackSetting', csrf, Admin.api.notification.removeSlackSetting)
```
lightnet328, line 29: Is it okay? I thought the naming convention of other routes is `resource.action`.
```suggestion
  router.post('/admin/notification/slackSetting.remove', csrf, Admin.api.notification.removeSlackSetting)
```
In other words, I think the above example follows other routes naming.

lib/controllers/admin.ts
```diff
@@ -127,12 +127,26 @@ export default (crowi: Crowi) => {
...
     } catch (err) {
       return res.json(ApiResponse.error(err.message))
     }
   }
 
+  actions.api.notification.removeSlackSetting = async function(req: Request, res: Response) {
+    try {
+      Config.deleteConfig('notification', 'slack:clientId')
+      Config.deleteConfig('notification', 'slack:clientSecret')
+      Config.deleteConfig('notification', 'slack:token')
```
lightnet328, line 141:
```suggestion
      await Promise.all([
        Config.deleteConfig('notification', 'slack:clientId'),
        Config.deleteConfig('notification', 'slack:clientSecret'),
        Config.deleteConfig('notification', 'slack:token'),
      ])
```
To catch errors, please wait promises.

sotarok: It should be. Thanks.

lib/routes/api/admin.ts
sotarok: OK, an if we do so, `/slackSetting` should also be `/slackSetting.add`, I'll fix it.

lightnet328: I see.

sotarok: fixed.

client/components/Admin/Share/SharePage.tsx
```diff
@@ -9,6 +9,7 @@ const SharePage: FC<{}> = () => {
   const [t] = useTranslation()
   const { crowi, loading, settingForm, fetchSettings } = useContext(AdminContext)
   const [alert, setAlert] = useState({ status: '', show: false, message: '' })
+  console.log('settingForm', settingForm)
```
lightnet328, line 12:
```suggestion
  console.log('settingForm', settingForm)
```

lib/service/config.ts
```diff
@@ -39,28 +49,72 @@ export default class Config {
 
   set(config) {
     // FIXME: Deep copy to avoid deleting itself.
-    config = JSON.parse(JSON.stringify(config))
+    const newConfig = { ...config }
```
lightnet328, line 52: Maybe this change will cause this problem. Would you like to undo this change and work on this ticket? We can make it so simple like the following if we worked on it.
https://github.com/crowi/crowi/pull/475

```js
set(config) {
  this.config = config
}
```

lib/models/config.ts
```diff
@@ -18,13 +18,53 @@ export interface ConfigDocument extends Document {
   value: string
 }
 
+export const registrationMode: Record<string, any> = {
+  [SECURITY_REGISTRATION_MODE_OPEN]: '公開 (だれでも登録可能)',
+  [SECURITY_REGISTRATION_MODE_RESTRICTED]: '制限 (登録完了には管理者の承認が必要)',
+  [SECURITY_REGISTRATION_MODE_CLOSED]: '非公開 (登録には管理者による招待が必要)',
+}
+
+export function isRequiredThirdPartyAuth(config) {
```
lightnet328, line 27: Just question, why did you move static method to just functions? I have a concern about mixing functions and static methods at the model layer.

lightnet328: BTW, I think it's better to add a type.

```
export function isRequiredThirdPartyAuth(config: Config) {
```

lib/service/config.ts
```diff
@@ -39,28 +49,72 @@ export default class Config {
...
     // FIXME: Treat as a mutable object always.
     //        We should always get config using `crowi.getConfig()` *just before* referencing config.
     for (const key of Object.keys(this.config)) {
       delete this.config[key]
     }
-    for (const key of Object.keys(config)) {
-      this.config[key] = config[key]
+    for (const key of Object.keys(newConfig)) {
+      this.config[key] = newConfig[key]
     }
```
lightnet328, line 60: memo:
We can make it more simple like the following unless considering the direct assignment of the config object. But it is okay working on this ticket.
https://github.com/crowi/crowi/issues/685

```js
set(config) {
  this.config = config
}
```

lib/service/config.ts
```diff
@@ -39,28 +49,72 @@ export default class Config {
...
+  }
+
+  /**
+   * To update the property of ConfigService,
+   * the config serice user (e.g. some controller, model, etc.) must call this update() method.
+   * Otherwise,
+   *  - the config memory cache would not be refreshed,
+   *  - the service like Slack, Mailer etc. would not be reloaded,
+   *  - the other server (in multi-server structure) would not be notified the config updated.
+   */
```
lightnet328, line 117: Nice👍

lib/models/config.ts
sotarok: > why did you move static method to just functions

IMO, we can make Config model's function simpler with separating static model functions and these simple utility function, 
And it could be imported as a single utility function without model instance from anywhere.

lightnet328: That makes sense. You meant you wanted to get some information from the config object like calling Config document object method without Config model's static method but you can't because it is a plain JavaScript object, didn't you?
I just had a concern about how to separate static method and plain functions in all models. However, I'm okay that making Config model's static methods for getting some information to be a function instead of making all static methods of all models to be a function as much as possible.

In the future, I think we can make a config object to be an entity class. Then, we don't have to call functions and dependent Config model.

PR #684, opened 2020-05-17
Fix/admin user list pager

sotarok:
## What

- Currently, all page list appears on pagination.
  - So if the number of user is quite large, the pages link overflowed outside of the screen like:
    - 
image
- Adjust to the original spec that assumed previous/next 2 pages appear around current page and show first/past link.
- To keep compatibility (for another pages using `Common/Pagination` component, it still works both with `pager` param and with `count` and `current` params.

##  Before

Screenshot 2020-05-17 11 55 19

### DIFF
--- a/client/components/Admin/User/UserPage.tsx
+++ b/client/components/Admin/User/UserPage.tsx
@@ -55,48 +55,49 @@ function useInviteUsers(crowi, fetchUsers, setFailure, clearStatus) {
 function useQuery(crowi) {
   const { search = '' } = crowi.location
   const { page: p = '', uq: q = '' } = queryString.parse(search)
-  const initialPage = Array.isArray(p) ? parseInt(p[0]) : p ? parseInt(p) : 0
+  const initialPage = Array.isArray(p) ? parseInt(p[0]) : p ? parseInt(p) : 1
   const initialQuery = (Array.isArray(q) ? q[0] : q) || ''
 
-  const [page, setPage] = useState(initialPage)
+  const [currentPage, setCurrentPage] = useState(initialPage)
   const [query, setQuery] = useState(initialQuery)
 
   return [
-    { page, query },
-    { setPage, setQuery },
+    { currentPage, query },
+    { setCurrentPage, setQuery },
   ] as const
 }
 
 function useFetchUsers(crowi, setFailure, clearStatus) {
   const [users, setUsers] = useState([])
-  const [{ page, query }, { setPage, setQuery }] = useQuery(crowi)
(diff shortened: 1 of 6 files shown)


client/components/Common/Pagination.tsx
```diff
@@ -1,47 +1,103 @@
...
+export interface Pager {
+  page: any
+  pagesCount: number
+  pages: number[]
+  total: number
+  previous: number | null
+  previousDots: boolean
+  next: number | null
+  nextDots: boolean
+}
```
lightnet328, line 14: Regarding using these types, I think the server-side pager object is for twig template, not for React components.
In other words, I'd like to make the pager type more simple when we remove `views/widget/pager`. But I have return pager object for React components when I migrated admin pages😇

BTW, Would you like to stop creating pagination from the pager object? I think the view itself that the new component renders is so great.

sotarok: Alright, it should be on React.
(In this PR, I didn't remove the pagination logic on server side though ...)

client/components/Admin/User/UserPage.tsx
```diff
@@ -55,45 +55,45 @@ function useInviteUsers(crowi, fetchUsers, setFailure, clearStatus) {
 function useQuery(crowi) {
   const { search = '' } = crowi.location
   const { page: p = '', uq: q = '' } = queryString.parse(search)
-  const initialPage = Array.isArray(p) ? parseInt(p[0]) : p ? parseInt(p) : 0
+  const initialPage = Array.isArray(p) ? parseInt(p[0]) : p ? parseInt(p) : 1
```
sotarok, line 58: the spec of `mongoose-paginate`, the page starts with `1`

client/components/Admin/User/UserPage.tsx
```diff
@@ -55,45 +55,45 @@ function useInviteUsers(crowi, fetchUsers, setFailure, clearStatus) {
...
     clearStatus(null)
 
     try {
-      const { users, pager } = await crowi.apiGet('/admin/users', { uq: search, page, ...options })
+      const { users, pager } = await crowi.apiGet('/admin/users', { uq: search, page: currentPage, ...options })
       setUsers(users)
-      setPage(pager.page)
-      setCount(pager.pagesCount)
+      setCurrentPage(pager.page)
+      setTotalPages(pager.pagesCount)
```
sotarok, line 83: It is a bit confusing that `count` means total item count or total page count, so changed to `totalPages`.
and `page` also a bit confusing. It is `currentPage`.

PR #694, opened 2020-06-22
Focus input in page create modal automatically

saitoeku3:
bug fix

Wait for the `Input` in `Modal` to be mounted using `setTimeout`.
This problem caused by reactstrap.

### DIFF
--- a/client/components/Modal/PageCreateModal.tsx
+++ b/client/components/Modal/PageCreateModal.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState, FC } from 'react'
+import React, { useCallback, useRef, useState, FC } from 'react'
 import { useTranslation, Trans } from 'react-i18next'
 import { Button, Form, Input, Label, Modal, ModalBody, ModalHeader, ModalProps } from 'reactstrap'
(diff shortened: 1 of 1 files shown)


client/components/Modal/PageCreateModal.tsx
```diff
@@ -88,6 +88,18 @@ const PageCreateModal: FC<Props> = ({ crowi, fade = false, toggle, ...modalProps
   const [portalName, setPotalName] = useState<string>(t('Memo'))
   const [pageName, setPageName] = useState('')
   const [underTreePath, setUnderTreePath] = useState(decodeURI(parentPath(currentPath)))
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  useEffect(() => {
+    if (!modalProps.isOpen) return
+
+    // Input is not mounted immediately after isOpen changed true
```
sotarok, line 96: Why don't you try `onOpen` callback?
https://reactstrap.github.io/components/modals/

```
// called when done transitioning in
 onOpened: PropTypes.func,
```

saitoeku3: Fixed!

PR #707, opened 2020-08-08
Implement auth settings form

lightnet328:
# WHAT

- Implement auth settings form

localhost_3000_admin_app(iPad Pro)

# WHY

- Auth settings form was removed accidentally when admin pages were migrated to React
  - https://github.com/crowi/crowi/pull/423


sotarok:
Code is okay 🙆‍♂️ 
But I'll check the behavior by this settings.

sotarok:
⚡ Found a bug, this may not directly related to this PR though ... an error occurred when 

- Set a domain to the allowing list of registering email address
- Access to the user settings page `/me`

then following error occurred.
 
image

PR #709, opened 2020-08-09
add user edit modal

lyohe:
## Overview

* added a modal client, route definition, form validator, model function and API for editing name and email address of users on admin console
* fixed some typo


lyohe:
😭

PR #714, opened 2020-08-22
Fix non existent user page

### LINKED ISSUE #272: /user/<nonexistent user> page can't be removed
1. Create /user/foo  from crowi command line tool (https://github.com/b4b4r07/crowi)
2. Open /user/foo from the web
3. The page can't be removed

saitoeku3:
## WHAT

- Ban to create user page manually
- Disable page create button when user page path
- Hide "Put Back" button on non existent user page in trash
- Make non existent user page deletable and movable

## WHY

https://github.com/crowi/crowi/issues/272

## SCREENSHOTS

Screen Shot 2020-08-22 at 15 32 56

Screen Shot 2020-08-22 at 15 34 01

Screen Shot 2020-08-22 at 15 33 22

## Issues

- Close #272

### DIFF
--- a/client/components/Modal/PageCreateModal.tsx
+++ b/client/components/Modal/PageCreateModal.tsx
@@ -6,18 +6,7 @@ import styled from 'styled-components'
 
 import { dark, gray, light } from '../../constants/colors'
 import Crowi from 'client/util/Crowi'
-
-const parentPath = (path: string) => {
-  if (path === '/') {
-    return path
-  }
-
-  if (path.match(/.+\/$/)) {
-    return path
-  }
-
-  return path + '/'
-}
+import { isUserPage, parentPath } from 'server/util/view'
 
 const TodayInputBox = styled.div`
   display: flex;
@@ -142,7 +131,9 @@ const PageCreateModal: FC<Props> = ({ crowi, fade = false, toggle, ...modalProps
                 onChange={(event: React.ChangeEvent<HTMLInputElement>) => setUnderTreePath(event.target.value)}
                 placeholder={t('Input page name')}
               />
-              <CreateButton color="primary">{t('Create')}</CreateButton>
+              <CreateButton color="primary" disabled={isUserPage(underTreePath)}>
+                {t('Create')}
+              </CreateButton>
             </Form>
           </>
         )}
--- a/lib/events/user.ts
+++ b/lib/events/user.ts
@@ -20,6 +20,7 @@ export default class UserEvent extends EventEmitter {
 
     if (page === null) {
       await this.createUserPage(userPagePath, user)
+      return
     }
 
     // User page created manually is already exists.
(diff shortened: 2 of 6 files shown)


client/components/Modal/PageCreateModal.tsx
```diff
@@ -142,7 +131,9 @@ const PageCreateModal: FC<Props> = ({ crowi, fade = false, toggle, ...modalProps
                 onChange={(event: React.ChangeEvent<HTMLInputElement>) => setUnderTreePath(event.target.value)}
                 placeholder={t('Input page name')}
               />
-              <CreateButton color="primary">{t('Create')}</CreateButton>
+              <CreateButton color="primary" disabled={isUserPage(underTreePath)}>
+                {t('Create')}
+              </CreateButton>
```
sotarok, line 136: Note: just disabled create button when the path is the user page

lib/events/user.ts
```diff
@@ -11,25 +13,35 @@ export default class UserEvent extends EventEmitter {
...
+    try {
+      const page = await Page.findPage(userPagePath, user, {}, false)
+
+      if (page === null) {
+        throw new Error()
+      }
+
+      // User page is already exists.
+      const renamedUserPagePath = `/tmp/user-${user.username}-${Date.now()}`
+      await Page.rename(page, renamedUserPagePath, user, {})
```
sotarok, line 29: 💬 We have to decide how to handle pages under the user page manually created.

- Stay exist on the original path (Move pages to another place manually if user want to move pages under the user page that might created by the other user) 
- Or ... also move all pages under the user page at the same time.

lib/events/user.ts
```diff
@@ -11,25 +13,35 @@ export default class UserEvent extends EventEmitter {
...
+      await this.createUserPage(userPagePath, user)
+    } catch {
+      await this.createUserPage(userPagePath, user)
```
sotarok, line 32: 💬 If you handle page existence with throwing `Error`, we won't be able to handle it separately from other types of errors 🙁 (e.g. failing page rename).
It would be better if handle the page existence just with control flow.

sotarok (commented):
Commented just at nits.

lib/events/user.ts
saitoeku3: Good suggestion!
I noticed `ignoreNotFound` option that `Page.findPage` return `null` instead of throwing `Error`.

lib/events/user.ts
saitoeku3: My opinion is "stay exist on the original path".

When moving all pages under the user page, links to moved pages in documents are broken.
I think the page should be changed at the user's discretion.

saitoeku3: @lightnet328 @sotarok Which is your opinion?

sotarok: > My opinion is "stay exist on the original path".

Agree with you!

PR #669, opened 2020-04-04
Migrate twig templates using single layout to JSX

lightnet328:
# WHAT

- Migrate twig templates using single layout
  - Search result page
  - Notification list page
  - API token page
  - Me page
  - Password page

# WHY

See #639

This PR depends on #642

PR #718, opened 2020-08-31
Fix error on share page for anonymous users

sotarok:
## What

- Error occurred on external share page
- This caused since #669 

image

## How to fixed it

- Move `views/page_share` to `server/pages/SharePage` 

## Note

- The page body is still rendered by CrowiRenderer called by jQuery, and also handle `secret-keyword-form-container` by React after it rendered.
- It would be the best if we implement render page body itself on server side, but to place a priority on fixing this problem, I just moved a code from `views/page_share` and fixed it. (Yes, it is a bit confusing though)


saitoeku3 (approved):
💯

PR #764, opened 2021-05-02
Fix: page `/admin/app` did not work on newly-installed crowi

kaz:
If some settings value is missing, `/admin/page` will be stopped due to errors.

See an image below:

スクリーンショット 2021-05-02 15 13 33

I tested with latest Firefox, but this error should be caused on other browsers.


sotarok:
Thank you!! ...and sorry for late response too 😢

PR #765, opened 2021-05-02
Fix: changing passwords fails

kaz:
- An input element for current password is missing on `/me/password` because `prop:hasPassword` is not passed to a component.
  - Due to this, users cannot change own password.
- A message `Please set a password` is unexpectedly shown for users who already has passwords.


sotarok:
Thank you and sorry for late 😢😢

PR #787, opened 2021-09-01
fix: error on admin page when securyt settings are empty

sotarok:
fix error when the settings are empty.


sotarok:
duplecated #764

PR #844, opened 2022-09-01
Fix es version to 7.13

sotarok:
On v1.8, Crowi supports ElasticSearch and its SaaS version (like Bonsai), and Bonsai uses OpenSearch in its backend.
Unfortunately, ElasticSearch and OpenSearch broke its relationship and from ES clietn library v7.14 and above hadn't supported OpenSearch.

We would support both ES itself and OpenSearch in v1.9~.
Currently, support both above with the current code base.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14150,9 +14150,9 @@
       "dev": true
     },
     "es7": {
-      "version": "npm:@elastic/elasticsearch@7.17.0",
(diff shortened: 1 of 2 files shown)


suzuki (approved):
> We would support both ES itself and OpenSearch in v1.9~.

LGTM

PR #851, opened 2022-10-13
Admin server info

suzuki:
## About

- I'm trying to change the requirements Node version of Crowi
- Sometimes, I get confused about which version of Node I'm running Crowi on
- So I want to add the Node version on the Admin screen

## Changing point

- `Node arch / version: x64 / v12.18.2`

スクリーンショット 2022-10-14 8 29 39


sotarok (approved):

sotarok:
Thanks!

PR #914, opened 2026-07-08
feat(i18n): add Korean (ko) translation

moduvoice:
## Motivation

This adds Korean (`ko`) translation support — to help Korean-speaking users use this
open-source project more comfortably in their own language.
(한국 사람들의 오픈소스 이용에 도움이 되게 하기 위해서 한글화 작업을 하였습니다.)

## Changes

- `packages/web/messages/ko.json` (new): full Korean translation of every key in `en.json` (1066 message keys).
- `packages/web/project.inlang/settings.json`: added `"ko"` to `locales` (`["ja", "en", "ko"]`) so paraglide compiles it as a live UI locale.
- `packages/web/src/components/locale-switcher.tsx`: added `ko: '한국어'` to `LOCALE_LABELS` so the pre-auth language switcher (login/register/etc.) offers Korean.
- The notification message templates (`notifications.message_one_user*`) follow the same per-locale reordering already used by the Japanese translation (see the comment in `packages/web/src/lib/notification-format.ts`), since a literal English word order reads unnaturally in Korean — this mirrors the existing `ja.json` pattern rather than introducing a new one.

**Intentionally out of scope:** the signed-in "Settings → Language" dropdown (`profile-form.tsx`) and the backend `Language` / `UserLanguage` Zod enums (`packages/api-contract`) are left untouched. That dropdown is backed by a persisted, server-validated `User.lang` field; widening it to accept `ko` would require backend schema/logic changes, which this translation-only PR does not include. Korean is fully usable via the paraglide locale switcher on the public/pre-auth pages and anywhere `m['...']()` renders.

## Testing

- Verified `ko.json` is valid JSON and has exactly the same key set as `en.json` (0 missing / 0 extra keys) via a Python diff script.
- Verified every `{placeholder}` token (e.g. `{email}`, `{count}`, `{path}`) in each English string is preserved verbatim in the corresponding Korean string (0 mismatches).
- Manually reviewed the translated strings for natural, consistent Korean UI phrasing (button/action verbs, error messages, admin copy), matching the tone of existing `ja.json`/`en.json` entries.
- Could not run `pnpm install` / `paraglide:compile` / `pnpm type-check` locally in this environment (repo pins `engines.node: 24.x`; local Node is 25.9.0), so the generated `@paraglide/runtime.js` `Locale` union could not be regenerated and compiled here. The change is additive and structurally identical to the existing `en`/`ja` entries (`Record<Locale, string>` gains a `ko` key), so it is expected to type-check cleanly once `pnpm paraglide:compile` regenerates the locale union — happy to fix up in CI if anything surfaces.


sotarok:
Thanks for taking the time to translate all 1,066 message keys into Korean, and for wiring it up properly (inlang settings + locale switcher included)!

Adding a new UI locale is a long-term commitment for the project rather than a one-time diff: every future feature adds new message keys, and a locale without an active maintainer falls behind quickly — which ends up being a worse experience than the English fallback. Since none of the current maintainers can review Korean text, we'd like to ask two things before deciding:

1. Are you (or your team) actually using Crowi? We'd love to hear about the use case.
2. Would you be willing to keep the Korean locale up to date as new strings land (roughly: respond to a ping when a release adds keys)?

If both are a yes, we're happy to move forward with review. If we don't hear back in a couple of weeks we'll close this for now — with no hard feelings, and we'd welcome it again once there's demand and a maintainer for the locale.

sotarok:
Closing due to no response to the follow-up request (asked 2026-07-12, no reply as of 2026-07-25). Happy to reopen if you'd like to pick this back up.

